PR #1, opened 2019-06-07
Testing copybara.

hlopko:


hlopko:
What's missing:
* buildkite presubmit

hlopko:
Since I'm not an owner I cannot know, but if you see a green merge button, we have to disable that :) The proper flow is let copybara import the PR into the internal repo, lgtm there, and let copybara export it to Github. Then we either remove admin rights from all human beings (leave only bazel-io account as admin), or at least show a big red warning that merging is not allowed for admins.

hlopko:
I'm curious if anybody got copybara email :) Not me.

PR #3, opened 2019-06-12
Rename java/defs.bzl to java/java.bzl 

iirina:
according to https://docs.bazel.build/versions/0.26.0/skylark/deploying.html


hlopko:
Actually, you already follow the updated policy (https://docs.bazel.build/versions/master/skylark/deploying.html) :P

Sorry for the trouble :/

iirina:
The actual name is correct according to the most updated version: https://docs.bazel.build/versions/master/skylark/deploying.html.

PR #8, opened 2019-07-12
Add Java toolchains & platforms.

iirina:
Add the required build_settings, config settings, constraint settings, constraint values, toolchain definitions and platforms that allow the majority of users to migrate their Java builds to the platform-based toolchain resolution. The PR implements the design described in [Migrate Java toolchains to platforms](https://docs.google.com/document/d/1dumNYb-3L8xswa2zjz7HizjhT05HIsEbQc2KuV3rdnk/edit?usp=sharing). Define:

* a build_setting for each of: java language version, javac version, JDK version, remote JDK version. * for each build_setting, an associated constraint_setting (language, javac, jdk, remote_jdk) and all the supported constraint_values (e.g. language_level/ javac_version/ jdk_version/remote_jdk_version{8, 9, 10, 11, 12})
* config_settings for each valid combination of build_setting (flag) values {language_version, javac_version} and similarly a config_setting for each supported jdk_version
* a toolchain of javac_toolchain_type for each combination of supported platform, language level and javac version (e.g. linux_javac10_language8_toolchain)
* a toolchain of java_runtime type for each supported platform and remote JDK version (e.g. linux_remote_jdk10_runtime_toolchain)
* a local_jdk toolchain that points to @locak_jdk//:jdk and is compatible with HOST_CONSTRAINTS exposed by @local_config_platforms//:constraints.bzl
* a platform for each supported platform, using  the config_settings define above to select the right constraint_values, depending on the chosen build_settings


iirina:
@hlopko Thanks for the review. I added more toolchains for java runtime, as well as for host javac and java runtime toolchains. PTAL.

java/toolchains/BUILD
```diff
@@ -0,0 +1,92 @@
...
+   toolchain_type = ":javac_toolchain_type"
+)
+
+toolchain(
+    name = "host_javac_darwin_toolchain",
+    exec_compatible_with = MACOS_CONSTRAINTS + HOST_JAVA_LANGUAGE_CONSTRAINTS,
+    target_compatible_with = MACOS_CONSTRAINTS + HOST_JAVA_LANGUAGE_CONSTRAINTS,
+    toolchain = "@remote_java_tools_darwin//:toolchain",
+    toolchain_type = ":javac_toolchain_type",
+)
```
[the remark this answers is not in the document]
hlopko: Note that platform with ("host", "jdk8", "java8") will match toolchain with ("jdk8, "java8"). It won't match toolchain e.g. with ("jdk8", "java8", "foo").

WORKSPACE
```diff
@@ -1 +1,70 @@
...
+    sha256 = "28989f78b1ce437c92dd27bb4943b2211ba4db916ccbb3aef83696a8f9b43724",
+    urls = [
+        "https://mirror.bazel.build/bazel_java_tools/releases/javac11/v3.0/java_tools_javac11_darwin-v3.0.zip",
+    ],
+)
+
+# The remote JDK 11 used by default in Bazel as host java runtime.
+maybe(
+    http_archive,
+    "host_jdk_linux",
```
hlopko: It's probably too late now, but wouldn't remote_jdk_linux be a better name for this repo?

java/toolchains/BUILD
```diff
@@ -0,0 +1,92 @@
...
+    name = "java_runtime_toolchain",
+    toolchain = "@local_jdk//:jdk",
+    toolchain_type = ":java_runtime_toolchain_type"
+)
+
+toolchain(
+    name = "host_java_runtime_linux_toolchain",
+    exec_compatible_with = LINUX_CONSTRAINTS + HOST_JAVA_RUNTIME_CONSTRAINTS,
+    target_compatible_with = LINUX_CONSTRAINTS + HOST_JAVA_RUNTIME_CONSTRAINTS,
+    toolchain = "@host_jdk_linux//:jdk",
```
hlopko: Here again, if we register java_runtime_toolchain before host_java_runtime_linux_toolchain, it will always be selected. Please correct my understanding if I'm wrong. I assume you wanted to give people the sane default of having java_runtime_toolchain available, and then give them a way to select downloaded java runtime when they want to use that or when they don't have the local_jdk installed at all?

Since user has to be explicit in their WORKSPACE file, we can tell them to either register java_runtime_toolchain, or host_java_runtime_linux_toolchain. That would be one way of choosing which one to use.

Alternative would be to actaully detect what java runtime is installed on the machine (I'd even inspect its version), and put exec_compatible_with constraint to java_runtime_toolchain, and also in the host platform. That tells Bazel that my host machine fulfills this constraints, and that java_runtime_toolchain can be selected. This way we can register both java_runtime_toolchain and host_java_runtime_linux_toolchain. 

Does this make any sense?

iirina: I think the terminology is a bit confusing. We previously had target and host tools. We now have target and execution platforms. The toolchains defined here should preserve the same state that Bazel currently defaults to:
- `@local_jdk//:jdk` for the target configuration
- Our remote jdk with a specific java version (11) `@host_jdk_linux//:jdk` for the host configuration

If the users need different toolchains they have to define and register them themselves.

After defining & registering these toolchains, my plan was to use

 ```
--platforms=@local_config_platform//:host \
--host_platform=@rules_java//java/platforms:host_java_platform
```

which should select `@rules_java//java/toolchains:java_runtime_toolchain` for the target config and `@rules_java//java/toolchains:host_java_runtime_X_toolchain` for the host config. From your comment I understand that this won't happen, even if I specify a host_platform? I would like to find a solution where we can register both toolchains.

Comment (iirina) [reply java/toolchains/BUILD]: Part of the output generated by `--toolchain_resolution_debug`:

```
Selected execution platform @rules_java//java/platforms:host_java_platform, type @rules_java//java/toolchains:java_runtime_toolchain_type -> toolchain @local_jdk//:jdk
Selected execution platform @rules_java//java/platforms:host_java_platform, type @rules_java//java/toolchains:java_runtime_toolchain_type -> toolchain @host_jdk_linux//:jdk
Selected execution platform @rules_java//java/platforms:host_java_platform, type @rules_java//java/toolchains:javac_toolchain_type -> toolchain @remote_java_tools_linux//:toolchain
```

which confuses mesurther, since it doesn't seem to try to find toolchains for the target platform, but it selects 2 runtime toolchains for the host platform. So I'm not understanding something. :)

Comment (iirina) [reply java/toolchains/BUILD]: I see. There is only one execution platform selected. The problem is that for Java we have the same target & host platform, but we need use another java version for building the tools in the host configuration. I can define the host toolchain using an alias and with no os/cpu constraints, e.g.:

```
toolchain(
    name = "host_javac_toolchain",
    exec_compatible_with = HOST_JAVA_LANGUAGE_CONSTRAINTS,
    target_compatible_with = HOST_JAVA_LANGUAGE_CONSTRAINTS,
    toolchain = select({
        "@bazel_tools//src/conditions:darwin": "@remote_java_tools_darwin//:toolchain",
        "@bazel_tools//src/conditions:windows": "@remote_java_tools_windows//:toolchain",
        "@bazel_tools//src/conditions:linux_x86_64": "@remote_java_tools_linux//:toolchain",
    }),
    toolchain_type = ":javac_toolchain_type"
)
```

Will try this after lunch.

iirina: Obsolete.

lberki:
Is this still relevant?

MikhailTymchukFT:
> Is this still relevant?

I double it. What is the current state of this?

hlopko:
Closing for lack of progress.

PR #10, opened 2019-07-30
Create repositories.bzl

Yannic:
This will help if we need to add dependencies or declare toolchains in
the future because users already have everything in their WORKSPACE

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,5 @@
 workspace(name = "rules_java")
 
-load("//java:repositories.bzl", "rules_java_dependencies", "rules_java_toolchains")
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -1 +1,5 @@
-# Empty WORKSPACE file.
+workspace(name = "rules_java")
+
+load("//java:repositories.bzl", "rules_java_dependencies", "rules_java_toolchains")
```
iirina: Can we use `@rules_java//java:repositories.bzl` here so this part can be easily copy-pasted?

Yannic: Sure, done!

iirina:
@Yannic I had to add module docstrings because buildifier was complaining about it in our internal presubmit, I hope it's ok!

PR #13, opened 2019-08-09
Fetch dependencies through the Bazel federation.

fweikert:
As a result of this commit, rules_java now contains the required bzl files to be tested as part of the Bazel federation.

### DIFF
--- a/java/repositories.bzl
+++ b/java/repositories.bzl
@@ -12,6 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# WARNING: This file only exists for backwards-compatibility.
+# rules_java uses the Bazel federation, so please add any new dependencies to
+# rules_java_deps() in
+# https://github.com/bazelbuild/bazel-federation/blob/master/repositories.bzl
+# Java-only third party dependencies can be added to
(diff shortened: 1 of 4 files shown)


fweikert:
Are there any blockers here?

java/repositories.bzl
```diff
@@ -12,6 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# WARNING: This file only exists for backwards-compatibility.
+# rules_java uses the Bazel federation, so please add any new dependencies to
```
iirina: Are the repositories in this file already defined in the federation? rules_java_deps() or rules_java_setup()?

fweikert: Yes:
https://github.com/bazelbuild/bazel-federation/blob/master/repositories.bzl#L169
https://github.com/bazelbuild/bazel-federation/blob/master/setup/rules_java.bzl#L3

fweikert: The repositories are here: https://github.com/bazelbuild/bazel-federation/blob/master/java_repositories.bzl

iirina: Thanks!

PR #26, opened 2020-01-16
Add a java_module rule

shs96c:
This adds a `java_module` rule to support building jars that can be used with the Java Platform Module System introduced in Java 9. It does a number of things:

 * Assumes that a module is circumscribed by the same boundary as a maven artefact (which is reasonable, and how systems such as gradle work already)
 * Infers as much as can be reasonably inferred using the JDK's own `jdeps` tool
 * Provides a mechanism for people to add additional information, for example preventing particular packages from being exported (to allow implementations to remain `public`, yet inaccessible to users of the Java Module), or to define the module as being open.
 * Infers name from maven coordinates (which is rapidly becoming a "best practice")

The inference of the basic structure of the module means that minimal changes need to be made when moving code around during refactoring, and for many users is correct enough to allow for production usage. Making this more explicit has lead to maintenance pain, and has occasionally left modules not being updated when they should be.

We use bytebuddy to emit the necessary bytecode, since the java compiler expects `module-info.java` to be compiled with all the classes that make up the module, and at the root of the tree. Given that this would prevent fine-grained dependencies within a java bazel build, and would be hard to make work in a mono-repo with a single java root, the approach taken seems reasonable.

This PR also has to work around known limitations in `singlejar`, most notably https://github.com/bazelbuild/bazel/issues/10507

I suspect that the handling of `JavaInfo` instances within this rule could be much improved, and I'd appreciate some advice on how to do that.


shs96c:
I know that the standard mechanism for importing libraries is to use `rules_jvm_external` and ask folks to add the dependencies. However, it would seem reasonable for `rules_jvm_external` to have a dependency on `rules_java`, and I'd rather not introduce a circular dependency. As such, I've checked the two jars needed by this rule into the tree.

If there's an alternative approach you'd like me to take, please let me know!

shs96c:
I've added my work email address to my gmail account. I think that'll cause the CLA bot to be happy.

shs96c:
@googlebot I signed it!

jin:
@iirina I think this rule fits in rules_java more than it does in rules_jvm_external, so I suggested @shs96c to open this PR here. Let me know if you feel otherwise!

johnynek:
I am glad to see someone working on this. Thanks for your work.

I have a sketch of a design here:

https://github.com/bazelbuild/rules_scala/issues/921

I don't fully understand how much of the problem this is supposed to solve. It looks like it is just making the jar itself manually, and not producing everything needed to publish to say sonatype (which requires sources, javadoc, and pgp signatures).

My design had a few features:
1. build a script/program that we run with bazel run which uses coursier to do the publishing (since coursier has some support and as adding more).
2. have the rule verify that all the dependencies of all the jars are other modules. Without this property, the pom/module will likely be broken for users.

I wanted to add the comments here to see if they are useful to you.

shs96c:
Once this is in place, I plan on making a PR for `rules_jvm_external` to get the `maven_artifact` rule from Selenium upstreamed: https://github.com/SeleniumHQ/selenium/blob/master/java/private/maven_artifacts.bzl That solves the problem of publishing to sonatype.

shs96c:
@googlebot I signed it!

shs96c:
Hurrah! That actually worked. I'm happy to make changes to this PR, so please LMK what would be good.

iirina:
Hi @shs96c, thanks for sending the PR. I didn't get to reviewing it yet, I'll try to get to it soon.

jin:
@iirina (and @cushon maybe?) - could you take a look at this PR please?

laurentlb:
@lberki

cushon:
This is @lberki's call, but I'm not sure this is the best location for these rules.

There is not currently a high level plan about how best to support modules in Bazel. I think that deserves a design review and discussion before starting implementation, and before making decisions that the stability guarantee will apply to.

Also the existing Java rules are all native rules, and are likely to stay that way in the near future, so `rules_java` currently just contains stubs for the other rules--none of the rule development is here.

I think it would be better to incubate these rules in a different repo. That way that can iterate more quickly, and they can still inform module support if/when it's added to the core ruleset.

shs96c:
@cushon, so I should move this to `rules_jvm_external`? I'm fine with that, though the module system is part of the native java platform, so I guess at some point, we want a `java_module` rule present in `rules_java` --- there's no scope for an `experimental` set of rules in this repo?

Would you also like me to start that design doc? I believe that Google aren't using modules yet, so perhaps it's best if someone outside of the company take the lead with that.... Is there a template to use?

jin:
I'm happy to help incubate the development of Java module support in rules_jvm_external, but I feel that the best place to put it is still rules_java, since r_j_e is for interacting with the external world.

Can we perhaps place this in an `experimental/` subdirectory and exclude that from the Copybara sync into Google?

comius:
Closing, because PR went stale and also because rules_jvm_external is a better location.

jin:
There's also https://github.com/bazel-contrib/rules_jvm, which is a better location than RJE.

PR #39, opened 2021-01-22
Fix java toolchain 4.1.0

comius:


toolchains/remote_java_repository.bzl
```diff
@@ -0,0 +1,88 @@
...
+    name = "prefix_version_setting",
+    values = {{"java_runtime_version": "{prefix}_{version}"}},
+    visibility = ["//visibility:private"],
+)
+config_setting(
+    name = "version_setting",
+    values = {{"java_runtime_version": "{version}"}},
+    visibility = ["//visibility:private"],
+)
+alias(
```
lberki, line 66: ...an alias that itself references to `config_setting` rules. That's... intricate. @gregestren will probably be proud of you :)

lberki (approved):

toolchains/remote_java_repository.bzl
gregestren: I first learned about this pattern from a Bazel user who was using it to make boolean logic on `select()`s work.  That inspired me to add some similar extensions to Skylib's `selects` library. I've found the pattern to be quite useful.

PR #54, opened 2022-04-08
Update version to 5.0.0

comius:


tetromino (changes requested):
Shouldn't this be 5.0.1?

The 5.0.0 tag was applied to commit 7a3c520737581f13691ad94a0f798a3518d869d1 where defs.bzl contained "4.1.0".

comius:
> Shouldn't this be 5.0.0.1?
> 
> The 5.0.0 tag was applied to commit [7a3c520](https://github.com/bazelbuild/rules_java/commit/7a3c520737581f13691ad94a0f798a3518d869d1) where defs.bzl contained "4.1.0".

The release tar contains no references to version (5.0.0) and there are no other but this version commit in between, so I think just properly releasing 5.0.0 is ok.

tetromino (dismissed):

PR #56, opened 2022-05-18
support for jdk 8 on M1

apesternikov:


java/repositories.bzl
```diff
@@ -112,6 +112,20 @@ def remote_jdk8_repos(name = ""):
         ],
         version = "8",
     )
+    maybe(
+        remote_java_repository,
+        name = "remote_jdk8_macos_aarch64",
+        target_compatible_with = [
+            "@platforms//os:macos",
+            "@platforms//cpu:aarch64",
```
katre, line 120: Please use `arm64` instead of `aarch64` for names here.

[`@platforms//cpu:aarch64`](https://github.com/bazelbuild/platforms/blob/fbd0d188dac49fbcab3d2876a2113507e6fc68e9/cpu/BUILD#L18) is an alias to [`@platforms//cpu:arm64`](https://github.com/bazelbuild/platforms/blob/fbd0d188dac49fbcab3d2876a2113507e6fc68e9/cpu/BUILD#L63).

katre (changes requested):

java/repositories.bzl
apesternikov: Done

katre:
LGTM, but someone on the Java rules team should also approve and handle merging.

java/repositories.bzl
cushon: @katre all of the other JDKs are using aarch64, e.g. `remote_jdk8_linux_aarch64` in this file. Can we keep the names consistent to make it easier to automate updates to these, and then separately rename all of them if that's an high priority cleanup?

katre: I hadn't realized that.

Yes, let's keep them in sync and clean up all at once. Sorry for the churn.

apesternikov: no worries, rolled back

apesternikov:
Question to maintainers: how to upload the reffered file to mirror.bazel.build ?

apesternikov:
any chance to move this forward?

comius (approved):

Issue #64, opened 2022-07-21
Question: How to find a JDK runtime with `master`?


jpsim:
I've been using rules_java 5.0.0 successfully, but I'll soon need to update to include recent changes in `master` to support the latest bazel 6.x rolling releases.

Prior to 338c4184064f77d3d6ca8eae2bc5a605f8ea0da4 the JDK runtime toolchain was being found correctly, but after that commit I get the following error:

```
ERROR: /.../BUILD:21:19: While resolving toolchains for target //{my-target}: no matching toolchains found for types @bazel_tools//tools/jdk:runtime_toolchain_type
```

<details>
<summary>Here's the complete error</summary>

```
INFO: Repository remotejdk11_macos_aarch64 instantiated at:
  /Users/jsimard/src/envoy-mobile/WORKSPACE:58:26: in <toplevel>
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/envoy_mobile/bazel/envoy_mobile_dependencies.bzl:53:24: in envoy_mobile_dependencies
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/envoy_mobile/bazel/envoy_mobile_dependencies.bzl:62:28: in kotlin_dependencies
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_java/java/repositories.bzl:469:23: in rules_java_dependencies
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_java/java/repositories.bzl:233:10: in remote_jdk11_repos
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_java/toolchains/remote_java_repository.bzl:48:17: in remote_java_repository
Repository rule http_archive defined at:
  /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/bazel_tools/tools/build_defs/repo/http.bzl:353:31: in <toplevel>
[...]

PR #69, opened 2022-08-18
Bump version to 5.1.0

comius:


Wyverald (approved):

jpsim:
Not sure how many people this update will break if others are hitting #64 as we are.

PR #74, opened 2022-11-04
Fix problems with release and bump version to 5.3.3

comius:
Adding MODULE.bazel into the release archive, fixing url to point to actual release url.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -8,6 +8,7 @@ filegroup(
         "AUTHORS",
         "BUILD",
         "LICENSE",
+        "MODULE.bazel",
         "//java:srcs",
         "//toolchains:srcs",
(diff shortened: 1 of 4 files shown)


meteorcloudy (approved):

BUILD
```diff
@@ -8,6 +8,7 @@ filegroup(
         "AUTHORS",
         "BUILD",
         "LICENSE",
+        "MODULE.bazel",
```
comius, line 11: @meteorcloudy Should I also add WORKSPACE and WORKSPACE.bzlmod to the release?

meteorcloudy: They are not needed when your project is used as a dependency, so probably not.

PR #75, opened 2022-11-09
Fix skylib dependency

comius:


meteorcloudy:
Is `bazel_skylib` a dev dependency or not? If it is, we probably want to keep it as dev dependency and not build all targets like `@rules_java//...`?

meteorcloudy:
Can you also add 
```
    "repository": [
        "github:bazelbuild/rules_java"
    ],
```
in https://github.com/bazelbuild/rules_java/blob/master/.bcr/metadata.template.json?

comius:
> Is `bazel_skylib` a dev dependency or not? If it is, we probably want to keep it as dev dependency and not build all targets like `@rules_java//...`?

It's used in `java/BUILD` and `java/private/BUILD`, so it's not a dev dependency.

comius:
> Can you also add
> 
> ```
>     "repository": [
>         "github:bazelbuild/rules_java"
>     ],
> ```
> 
> in https://github.com/bazelbuild/rules_java/blob/master/.bcr/metadata.template.json?

Done

meteorcloudy (approved):

comius:
> > Can you also add
> > ```
> >     "repository": [
> >         "github:bazelbuild/rules_java"
> >     ],
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in https://github.com/bazelbuild/rules_java/blob/master/.bcr/metadata.template.json?
> 
> Done

I think this has broken bcr publish app. :(

PR #77, opened 2022-11-10
Fix failures and bump version to 5.3.5

comius:
- bumped rules_cc to 0.0.2, to be consistent with bazel_tools
 - removed all @// labels
 - tagged some targets in //toolchains/... as manual, so that `build //toolchains/... wouldn't break
 - build still breaks on some non-existent files in java_tools_prebuilt repository; those are never selected though
 - thus changed testing from @rules_java//... to @rules_java//java/...
 - bumped version to 5.3.5


meteorcloudy (approved):
Awesome!

Issue #70, opened 2022-09-06
Using default_java_toolchain, etc from //toolchains as opposed to @bazel_tools doesn't work


gmishkin:
get errors like 

```
ERROR: /Users/geoff/code/mms/server/BUILD.bazel:592:23: no such package 'toolchains': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /Users/geoff/code/mms/toolchains and referenced by '//server:java_toolchain_config'
ERROR: /Users/geoff/code/mms/server/BUILD.bazel:592:23: no such package 'toolchains': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /Users/geoff/code/mms/toolchains and referenced by '//server:java_toolchain_config'
ERROR: /Users/geoff/code/mms/server/BUILD.bazel:592:23: no such package 'toolchains': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /Users/geoff/code/mms/toolchains and referenced by '//server:java_toolchain_config'
ERROR: /Users/geoff/code/mms/server/BUILD.bazel:592:23: no such package 'toolchains': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /Users/geoff/code/mms/toolchains and referenced by '//server:java_toolchain_config'
```

I assume it's because all the labels in default_java_toolchain.bzl like "//toolchains:singlejar" need to be wrapped in `Label()` or something so that they're relative to @rules_java?

gmishkin:
Just as a heads up, this is an issue now when using default_java_toolchain from `@bazel_tools` on recent versions of rules_java, because rules_java now points to a version of remote_java_tools that no longer provides a `java_compiler_jar`, but bazel_tools' default_java_toolchain tries to patch the java module with `@remote_java_tools//:java_compiler_jar` which no longer exists.

Wyverald:
Could we do a rules_java release for this?

SimonVleugels:
@comius, When can we expect a new rules_java release? This is a blocking issue for us and we are trying to avoid using the master branch.

Wyverald:
To close the loop: there is a new release now https://registry.bazel.build/modules/rules_java

PR #92, opened 2023-03-15
Apply Label() wherever appropriate

Wyverald:
The current `default_java_toolchain` macro uses bare string literals for some "implicit dependencies", which means that those labels are resolved relative to the package being built, not the current .bzl file. This patch fixes that by using `Label()` wherever appropriate.

https://github.com/bazelbuild/rules_java/pull/85 already fixed some of these, and this patch fixes some more.

Fixes https://github.com/bazelbuild/rules_java/issues/70

### DIFF
--- a/toolchains/local_java_repository.bzl
+++ b/toolchains/local_java_repository.bzl
@@ -93,7 +93,7 @@ def local_java_runtime(name, java_home, version, runtime_name = None, visibility
     native.toolchain(
         name = "runtime_toolchain_definition",
         target_settings = [":%s_settings_alias" % name],
-        toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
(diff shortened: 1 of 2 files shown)


toolchains/local_java_repository.bzl
```diff
@@ -93,7 +93,7 @@ def local_java_runtime(name, java_home, version, runtime_name = None, visibility
     native.toolchain(
         name = "runtime_toolchain_definition",
         target_settings = [":%s_settings_alias" % name],
-        toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
+        toolchain_type = Label("@bazel_tools//tools/jdk:runtime_toolchain_type"),
         toolchain = runtime_name,
```
comius, line 97: I guess we don't need to add Label down in the line 230, becuase that's just a regular target in a BUILD file.

comius (approved):

PR #100, opened 2023-03-25
Add remote_jdk19_repos to rules_java_dependencies

BoleynSu:


hvadehra:
I think we'd only like to register the LTS versions by default. One can call the register methods for other versions if needed.

PR #101, opened 2023-04-03
Fix --incompatible_disallow_empty_glob

keith:
This is the same as https://github.com/bazelbuild/bazel/commit/c2ddbd1954af5baab63b93f2b055a410a27832c8


comius (approved):

keith:
@comius can you help land?

PR #107, opened 2023-05-02
Update java_tools to v12.2

keertk:
https://github.com/bazelbuild/java_tools/issues/70


keertk:
I guess this is no longer needed since we now have https://github.com/bazelbuild/rules_java/pull/108

PR #110, opened 2023-05-16
Sync java_tools, remote JDKs, and java toolchains in rules_java with Bazel HEAD

meteorcloudy:
Context: https://github.com/bazelbuild/bazel/discussions/18373
Preparing for using rules_java as the source of truth for all java toolchain related definitions (https://github.com/bazelbuild/bazel/pull/18423)


meteorcloudy:
@hvadehra @cushon This PR is ready for review, can you please take a look?

meteorcloudy:
FYI @keertk

meteorcloudy:
https://github.com/bazelbuild/bazel/pull/18423 is now blocking on this PR

hvadehra (approved):

meteorcloudy:
OK, this is causing many downstream breakages: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3066#_

The reason is that the default rules_java is defined in WORKSPACE suffix, which will be overridden by the rules_java version introduced in the user's WORKSPACE file. If that rules_java version is too old, loading JDK repos and toolchains will fail. I think we'll need to figure out a way to make sure the latest version of rules_java is used in WORKSPACE.

And this is not a problem with Bzlmod enabled because we use MVS to resolve Bazel module dependencies.

PR #114, opened 2023-06-09
Obtain bootstrap class path from dedicated runtime toolchain type

fmeum:
This decouples the concepts of a Java runtime used to run on the target platform from the Java runtime that provides the bootstrap class path used during compilation.

The former needs constraints on the target platform, whereas the latter should not have any constraints on the target platform to allow for cross-compilation to target platforms for which a JDK runtime is not available (e.g. Android).

Work towards bazelbuild/bazel#17085
Work towards #64
Split off from bazelbuild/bazel#18262


hvadehra (dismissed):
LGTM, but I'd like for @cushon to also have a look

cushon (dismissed):

fmeum:
@cushon @hvadehra I noticed that I forgot to register the toolchains in `MODULE.bazel` and added a new commit to do this.

hvadehra (dismissed):

fmeum:
@hvadehra @cushon Friendly ping. 

I pushed a commit to fix CI after the merge of https://github.com/bazelbuild/bazel/commit/8715e9ae6f5c2fd42607018b3adaad14bf601a1c. Let me know whether you would want me to handle this differently.

cushon (dismissed):

hvadehra (approved):

fmeum:
Had to push another commit to fix a breakage that CI didn't catch: I forgot to add the `register_toolchains` calls for the bootstrap runtime toolchains to the WORKSPACE macro.

Do you see a way to have CI catch such issues, for example by including a pipeline that runs Bazel@HEAD with an `--override_repository` for `@rules_java_builtin`?

CC @meteorcloudy in case you have an idea

meteorcloudy:
I do feel the test coverage for rules_java is quite limited, I think we can:
- add more examples under https://github.com/bazelbuild/rules_java/tree/master/examples/hello_world to cover more real world use cases. With WORKSPACE, you can use `--override_repository` to override `@rules_java_builtin` as Fabian suggested, with Bzlmod, you can use `local_path_override`. Then you can configure which Bazel version is used in the [presubmit.yml](https://github.com/bazelbuild/rules_java/blob/master/.bazelci/presubmit.yml) file.
- Or consider using https://github.com/bazel-contrib/rules_bazel_integration_test to add more integration tests?

I would suggest to look into `rules_bazel_integration_test` first to see if fits the need. ;)

fmeum:
@hvadehra I think so, thanks!

fmeum:
@hvadehra Could you create a release that I can then update Bazel to?

PR #126, opened 2023-08-10
Reference `platformclasspath` in `@bazel_tools`

fmeum:
The `platformclasspath` rule in `@rules_java` always requires a bootstrap Java runtime toolchain to be registered, but whether the commonly used toolchain macros in `@bazel_tools` register this toolchain depends on the version of Bazel that is running the build and the version of `rules_java` it bundles as `@rules_java_builtin`.

In order to support all Bazel versions, reference the `platformclasspath` in `@bazel_tools`, which in turn references the corresponding target in `@rules_java_builtin`, which may or may not be the current version of this code.


fmeum:
@hvadehra

PR #127, opened 2023-08-17
Fix docs on `local_java_repository`

fmeum:
The previously added recommendation to register the `:all` pseudo-target altered the previous behavior of the macro by also registering the compilation toolchain.


fmeum:
CC @hvadehra 

I have to say that I find `{local,remote}_java_repository` very confusing. The fact that these macros have differ in whether they define a compilation toolchain and use different suffixes for toolchain targets makes it really hard to keep track of what toolchains are actually ending up registered. Maybe there could be a "v2" at some point that organizes this better and makes runtime and compilation toolchain definition logically separate?

PR #133, opened 2023-09-22
Set `java` on remote JDKs

fmeum:
If the `java` attribute on `java_runtime` isn't set, Bazel guesses it based on the host platform, which can be different from the execution platform.

https://github.com/bazelbuild/bazel/blob/04a05677e4714434ec196c45a4ccbd8e0ef2f0ff/src/main/java/com/google/devtools/build/lib/rules/java/JavaRuntime.java#L46

Work towards https://github.com/bazelbuild/bazel/issues/19587

### DIFF
--- a/toolchains/jdk_build_file.bzl
+++ b/toolchains/jdk_build_file.bzl
@@ -83,6 +83,13 @@ java_runtime(
         ":jdk-lib",
         ":jre",
     ],
+    # Provide the 'java` binary explicitly so that the correct path is used by
+    # Bazel even when the host platform differs from the execution platform.
+    # Exactly one of the two globs will be empty depending on the host platform.
+    # When --incompatible_disallow_empty_glob is enabled, each individual empty
+    # glob will fail without allow_empty = True, even if the overall result is
+    # non-empty.
+    java = glob(["bin/java.exe", "bin/java"], allow_empty = True)[0],
     version = {RUNTIME_VERSION},
 )
 """


fmeum:
CC @tjgq @uebelandre

cushon (approved):

tjgq:
@cushon @hvadehra WDYT about this approach?

toolchains/jdk_build_file.bzl
```diff
@@ -83,6 +83,9 @@ java_runtime(
         ":jdk-lib",
         ":jre",
     ],
+    # Provide the 'java` binary explicitly so that the correct path is used by
+    # Bazel even when the host platform differs from the execution platform.
+    java = glob(["bin/java.exe", "bin/java"])[0],
```
tjgq, line 88: Should we also:

1. Document that `java_runtime.java` should (must?) be explicitly set, otherwise the default will be a guess based on the host platform?
1. Update the TODO in `JavaRuntime.java` accordingly?

tjgq: (Whoops, I just noticed that this is not the main Bazel repo, but the comments still apply outside of this PR)

cushon: > 1. Document that `java_runtime.java` should (must?) be explicitly set, otherwise the default will be a guess based on the host platform?
> 2. Update the TODO in `JavaRuntime.java` accordingly?

Documenting it as recommended sounds good to me.

I wonder if there's a path to making it mandatory and removing the logic for inferring it, but I'm not sure what the cleanup for that would look like.

cushon:
> @cushon @hvadehra WDYT about this approach?

LGTM

toolchains/jdk_build_file.bzl
UebelAndre, line 88: It'd be nice if this worked with https://github.com/bazelbuild/bazel/issues/8195
```suggestion
    java = glob(["bin/java.exe", "bin/java"], allow_empty = True)[0],
```

UebelAndre (changes requested):

toolchains/jdk_build_file.bzl
fmeum: Is every individual pattern checked for emptiness?

fmeum: Looks like it.

cushon (dismissed):

toolchains/jdk_build_file.bzl
UebelAndre: I also filed https://github.com/bazelbuild/rules_java/issues/134

UebelAndre:
@cushon @comius @tjgq Friendly ping, Is this ready to merge?

toolchains/jdk_build_file.bzl
```diff
@@ -83,6 +83,9 @@ java_runtime(
         ":jdk-lib",
         ":jre",
     ],
+    # Provide the 'java` binary explicitly so that the correct path is used by
+    # Bazel even when the host platform differs from the execution platform.
+    java = glob(["bin/java.exe", "bin/java"], allow_empty = True)[0],
```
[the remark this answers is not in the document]
fmeum: I added a comment to explain why this is needed.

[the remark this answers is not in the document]
hvadehra: Thanks!

hvadehra (approved):

PR #142, opened 2023-10-17
Revert "Reference `platformclasspath` in `@bazel_tools`"

fmeum:
This reverts commit 54e89daeacd9728f45b4c16b3834e16d5f40e11c.

Going through `@bazel_tools` to reference the bootstrap classpath is no longer needed with Bazel 7, which always uses toolchain macros that define the required bootstrap runtime toolchains.

Since the original problem didn't arise with Bzlmod, where the Java toolchains have always come from rules_java rather than bazel_tools, this fixes https://github.com/bazelbuild/bazel/issues/19837 for Bazel itself by allowing it to use the new bootstrap classpath rule while still being built with Bazel 6.4.0.

### DIFF
--- a/toolchains/default_java_toolchain.bzl
+++ b/toolchains/default_java_toolchain.bzl
@@ -86,7 +86,7 @@ _BASE_TOOLCHAIN_CONFIGURATION = dict(
     singlejar = [Label("//toolchains:singlejar")],
     # Code to enumerate target JVM boot classpath uses host JVM. Because
     # java_runtime-s are involved, its implementation is in @bazel_tools.
-    bootclasspath = [Label("@bazel_tools//tools/jdk:platformclasspath")],
+    bootclasspath = [Label("//toolchains:platformclasspath")],
     source_version = "8",
     target_version = "8",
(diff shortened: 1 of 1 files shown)


fmeum:
CC @hvadehra

fmeum:
Is the next release planned to be `7.0.0`? This commit would break compatibility with Bazel 6.

hvadehra:
> 
Is the next release planned to be 7.0.0? This commit would break compatibility with Bazel 6.

Yes, there are also another breaking change in the next release. 

Is this change needed for Bazel 7?

fmeum:
> Is this change needed for Bazel 7?

It is needed to fix https://github.com/bazelbuild/bazel/issues/19837, which currently blocks https://github.com/bazelbuild/bazel/pull/18262 and that would be great to get into Bazel 7. As an alternative, we could also cherry-pick https://github.com/bazelbuild/bazel/commit/149cbce5418fadaac329dd007b1411407fea55c3 into 6.4.0, but I don't know whether that is 100% safe.

PR #164, opened 2023-12-21
Remove explicit Bazel version from `presubmit.yml`

fmeum:
The BCR should be free to set this version and now uses 7.0.0 by default, which matches the compatibility of this ruleset.


fmeum:
@hvadehra

hvadehra (approved):

PR #171, opened 2024-02-12
Check local version detection result

### LINKED ISSUE #170: Broken local Java version detection
The current local Java version detection in [_detect_java_version](https://github.com/bazelbuild/rules_java/blob/master/toolchains/local_java_repository.bzl#L20) may return `None`, which in turn breaks further processing in [_local_java_repository_impl](https://github.com/bazelbuild/rules_java/blob/master/toolchains/local_java_repository.bzl#L206).

I encountered this problem due to a broken local Java installation, which was not able to execute `java -XshowSettings:properties` to begin with.

Error message:
```
ERROR: <builtin>: fetching _local_java_repository_rule rule //:rules_java~7.3.2~toolchains~local_jdk: Traceback (most recent call last):
    File "/private/var/tmp/_bazel_administrator/02bc2c2b51523be4b85c9cb7f6299f33/external/rules_java~7.3.2/toolchains/local_java_repository.bzl", line 217, column 72, in _local_java_repository_impl
        build_file = build_file.format(RUNTIME_VERSION = version if version.isdigit() else "0")
Error: 'NoneType' value has no field or method 'isdigit'
```

bertschneider:
In case the local Java version detection fails, e.g. due to a corrupt installation, skip further processing, as it results in a type error.

Closes #170

### DIFF
--- a/toolchains/local_java_repository.bzl
+++ b/toolchains/local_java_repository.bzl
@@ -191,12 +191,17 @@ def _local_java_repository_impl(repository_ctx):
 
     if not java_bin.exists:
         # Java binary does not exist
+        message = ("Cannot find Java binary {java_binary} in {java_home}; either correct your JAVA_HOME, " +
+                   "PATH or specify Java from remote repository (e.g. " +
+                   "--java_runtime_version=remotejdk_11)")
         repository_ctx.file(
             "BUILD.bazel",
-            _NOJDK_BUILD_TPL.format(
+            _AUTO_CONFIG_ERROR_BUILD_TPL.format(
                 local_jdk = local_java_runtime_name,
-                java_binary = _with_os_extension(repository_ctx, "bin/java"),
-                java_home = java_home,
+                message = repr(message.format(
+                    java_binary = _with_os_extension(repository_ctx, "bin/java"),
+                    java_home = java_home,
+                )),
             ),
             False,
         )
@@ -204,6 +209,21 @@ def _local_java_repository_impl(repository_ctx):
 
     # Detect version
     version = repository_ctx.attr.version if repository_ctx.attr.version != "" else _detect_java_version(repository_ctx, java_bin)
+    if version == None:
+        # Java version could not be detected
+        message = "Cannot detect Java version of {java_binary} in {java_home}; make sure it points to a valid Java executable"
+        repository_ctx.file(
+            "BUILD.bazel",
+            _AUTO_CONFIG_ERROR_BUILD_TPL.format(
+                local_jdk = local_java_runtime_name,
+                message = repr(message.format(
+                    java_binary = _with_os_extension(repository_ctx, "bin/java"),
+                    java_home = java_home,
+                )),
+            ),
+            False,
+        )
+        return
 
     # Prepare BUILD file using "local_java_runtime" macro
     if repository_ctx.attr.build_file_content and repository_ctx.attr.build_file:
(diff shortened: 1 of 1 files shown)


bertschneider:
As the Java installation was detected via `JAVA_HOME`, I'm not sure if skipping it or raising some kind of error would be the better approach.

bertschneider:
cc @fmeum

toolchains/local_java_repository.bzl
```diff
@@ -204,6 +204,8 @@ def _local_java_repository_impl(repository_ctx):
 
     # Detect version
     version = repository_ctx.attr.version if repository_ctx.attr.version != "" else _detect_java_version(repository_ctx, java_bin)
+    if version == None:
+        return
```
fmeum, line 208: Since this happens in the implementation function of a repository rule, we need to generate something or the user will just get errors about missing targets. Could you do something similar to line 194 here?

bertschneider: I used the template from line `194` to indicate an error. The output is still somewhat verbose, though.

```
ERROR: /home/norbert/.cache/bazel/_bazel_norbert/88cb78e88cb9acdaa4bfce98a60561bf/external/rules_java~7.3.2~toolchains~local_jdk/BUILD.bazel:2:10: in fail_rule rule @@rules_java~7.3.2~toolchains~local_jdk//:jdk:
Traceback (most recent call last):
        File "/home/norbert/.cache/bazel/_bazel_norbert/88cb78e88cb9acdaa4bfce98a60561bf/external/rules_java~7.3.2/toolchains/fail_rule.bzl", line 19, column 13, in _fail_rule_impl
                fail("%s %s" % (ctx.attr.header, ctx.attr.message))
Error in fail: Auto-Configuration Error: Cannot detect Java version of bin/java in /home/norbert/.asdf/installs/java/openjdk-21; make sure it points to a valid Java executable
ERROR: /home/norbert/.cache/bazel/_bazel_norbert/88cb78e88cb9acdaa4bfce98a60561bf/external/rules_java~7.3.2~toolchains~local_jdk/BUILD.bazel:2:10: Analysis of target '@@rules_java~7.3.2~toolchains~local_jdk//:jdk' failed
```

toolchains/local_java_repository.bzl
```diff
@@ -239,14 +259,12 @@ local_java_runtime(
...
-_NOJDK_BUILD_TPL = '''load("@rules_java//toolchains:fail_rule.bzl", "fail_rule")
+# Build file template, when JDK could not be detected
+_AUTO_CONFIG_ERROR_BUILD_TPL = '''load("@rules_java//toolchains:fail_rule.bzl", "fail_rule")
 fail_rule(
    name = "jdk",
    header = "Auto-Configuration Error:",
-   message = ("Cannot find Java binary {java_binary} in {java_home}; either correct your JAVA_HOME, " +
-          "PATH or specify Java from remote repository (e.g. " +
-          "--java_runtime_version=remotejdk_11)")
+   message = "{message}"
```
fmeum, line 267: This is problematic if someone starts using double quotes in the message. Could you remove the quotes from the temple and instead pass `repr(...)` to the `format` call?

fmeum:
Looks like buildifier complains, but the other CI failure seems unrelated.

toolchains/local_java_repository.bzl
bertschneider: Ah, I see, good point! Changed.

bertschneider:
Do you have an idea what's causing the failed checks?

fmeum:
> Do you have an idea what's causing the failed checks?

Looks like this is coming from the new test setup:
https://github.com/bazelbuild/rules_java/blob/4f46f3a530c6fa0accb0d885bf2bd39de6c43e5d/.bazel_checkout/setup.sh#L5-L9

cc @hvadehra

fmeum (approved):

bertschneider:
Should the PR be merged regardless?

hvadehra:
Should be fixed after https://github.com/bazelbuild/rules_java/commit/88bb3ada19a2f33a51c62e6fc04193e7c0e15994. Could you please rebase on master?

hvadehra (approved):

bertschneider:
> Could you please rebase on master?

Done. Fixed the failed checks, thx!

toolchains/local_java_repository.bzl
```diff
@@ -191,19 +191,39 @@ def _local_java_repository_impl(repository_ctx):
...
             False,
         )
         return
 
     # Detect version
     version = repository_ctx.attr.version if repository_ctx.attr.version != "" else _detect_java_version(repository_ctx, java_bin)
+    if version == None:
+        # Java version could not be detected
+        message = "Cannot detect Java version of {java_binary} in {java_home}; make sure it points to a valid Java executable"
+        repository_ctx.file(
```
cushon, line 215: It might be worth extractor a helper function to share this with the copy above, I think the only different here is the `message`?

PR #174, opened 2024-03-08
Rename `java/modules` to `java/common`

fmeum:
The term "modules" is easily confused with "Bazel modules" from Bzlmod.


fmeum:
CC @hvadehra @comius

fmeum:
We discussed this on the ruleset structure design doc, where `common` was the preferred alternative.

hvadehra (approved):

PR #181, opened 2024-04-02
Register Java 8 runtimes with Bzlmod

fmeum:
These runtimes were only registered when using WORKSPACE.

Work towards bazelbuild/bazel#21769


fmeum:
CC @meteorcloudy

meteorcloudy (approved):

PR #182, opened 2024-04-02
Add `incompatible_language_version_bootclasspath` flag

fmeum:
With `--@rules_java//java:incompatible_language_version_bootclasspath`, the bootclasspath used for Java compilation is now determined based on the numeric version specified in `--java_language_version` and the type specified in `--java_runtime_version` rather than just using the target runtime.

For example, with `--java_language_version=8` and
`--java_runtime_version=remotejdk_21`, the bootclasspath would be extracted from `remotejdk_8`.

For unversioned runtime versions such as `local_jdk`, the behavior doesn't change.

If a matching runtime is not available, analysis fails with a customized error message explaining the various options to the user.

Work towards bazelbuild/bazel#21769


fmeum:
Sounds good! I resolved the conflicts.

hvadehra (commented):
Mostly LGTM, other than the minor organizational comments. 

Can we please add some rudimentary tests for this?

fmeum:
I added analysis tests and addressed the comments, please take a look.

PR #183, opened 2024-04-03
Add JDK 8 `tools.jar` contents to bootclasspath

fmeum:
Work towards https://github.com/bazelbuild/bazel/issues/21771


cushon (approved):

fmeum:
In terms of knobs, as it's essentially a portability issue, maybe this could become a new errorprone check that could then be configured on the toolchain/package/target level as usual?

PR #187, opened 2024-05-12
Mark `toolchains` extension as reproducible

fmeum:
This avoids an unnecessary and verbose lockfile entry for rules_java users. Uses `bazel_skylib`'s `modules.as_extension` to automatically detect the availability of the `reproducible` feature.


fmeum:
@hvadehra If it doesn't delay anything, could this be included in 7.6.0?

hvadehra (approved):

PR #191, opened 2024-05-17
Do not enable `use_all_repos` for toolchains extension

fmeum:
The toolchain extension isn't internal to rules_java as users have to `use_extension` it to get a reference to `local_jdk`. They should not see a warning about not importing all the other repos, which comes with `modules.as_extension` due to its use of `use_all_repos`. Instead, just declare the extension as reproducible.


hvadehra (approved):

fmeum:
@hvadehra Sorry for messing this up, I noticed it when I saw the warning pop up on Bazel itself. As it would make `bazel mod tidy` behave weirdly, we might want a 7.6.1 release.

PR #195, opened 2024-06-21
Fix path mapping logic in `bootclasspath` for generated files

fmeum:
If the files comprising the Java runtime are not source files, the path to the Java home needs to be path mapped.

Work towards bazelbuild/bazel#22366


fmeum:
@hvadehra

hvadehra (approved):

Issue #199, opened 2024-07-22
Release: rules_java 7.7.3


comius:
**Link to relevant issue and/or commit:**
https://github.com/bazelbuild/rules_java/commit/7517ac67831e286117ed85722810fc1407fa01a3

**Other details:**
I need this update to support java_proto_library in protobuf repo.

cc @bazelbuild/triage

keertk:
Done - 
https://github.com/bazelbuild/rules_java/releases/tag/7.8.0
https://github.com/bazelbuild/bazel-central-registry/pull/2470

keertk:
Released as 7.8.0 - 
https://github.com/bazelbuild/rules_java/releases/tag/7.8.0
https://github.com/bazelbuild/bazel-central-registry/pull/2471

PR #201, opened 2024-07-22
Update version to 7.7.3

keertk:
Recreate https://github.com/bazelbuild/rules_java/issues/199 as 7.7.3 instead of 7.8.0


keertk:
(not required. released as 7.8.0)

PR #206, opened 2024-08-23
Wire up `oneversion` in Java toolchain configurations

fmeum:


fmeum:
@cushon Since `--experimental_one_version_enforcement` is `OFF` by default, I think that it won't hurt to register it in the default toolchain configuration, even with no way to allowlist existing targets yet.

cushon (approved):

fmeum:
@keertk After this has been merged, could we get a new java_tools/rules_java release to pick this up together with the turbine update and https://github.com/bazelbuild/bazel/commit/407ed1dd17a0aa442816ac41efb6d81d8cf14ae9?

keertk:
@fmeum sure, tracked here: https://github.com/bazelbuild/java_tools/issues/91

MODULE.bazel
```diff
@@ -14,6 +14,7 @@ bazel_dep(name = "bazel_skylib", version = "1.6.1")
 # Required by @remote_java_tools, which is loaded via module extension.
 bazel_dep(name = "rules_proto", version = "4.0.0")
 bazel_dep(name = "rules_license", version = "0.0.3")
+bazel_dep(name = "abseil-cpp", version = "20240116.2", repo_name = "com_google_absl")
```
hvadehra, line 17: @fmeum Why was this new dep required in this change?

fmeum: This is a dep of oneversion when built from source.

hvadehra: I see, thanks.

Issue #211, opened 2024-09-03
Release: rules_java 7.10.0


comius:
**Link to relevant issue and/or commit:**

This version is now compatible with Bazel 6.2.0.
https://github.com/bazelbuild/rules_java/commit/30ecf3ff6ee8f30b4df505d9d3bde5bb1c25690b

**Other details:**
Please also update the version in java/defs.bzl

cc @bazelbuild/triage

keertk:
https://github.com/bazelbuild/rules_java/releases/tag/7.10.0
https://github.com/bazelbuild/bazel-central-registry/pull/2716

PR #212, opened 2024-09-03
Release rules_java v7.10.0

keertk:
https://github.com/bazelbuild/rules_java/issues/211


comius (changes requested):
Can you also change https://github.com/bazelbuild/rules_java/blob/master/java/defs.bzl#L28?

PR #238, opened 2024-10-30
Add the repo rule `http_jar`

Wyverald:
The repo rule `http_jar` currently lives in `@bazel_tools`, which results in an implicit dependency on `rules_java` because it creates a repo with a BUILD file that uses `java_import`. So this repo rule really belongs in `rules_java`.

This PR copies it over from `@bazel_tools//tools/build_defs/repo:http.bzl` with minimal edits.


Wyverald:
Marked as draft right now as I can't figure out how to add tests for this.

fmeum:
This would unblock https://github.com/bazelbuild/rules_java/pull/239, which currently fails because using `http_jar` from `bazel_tools` in `rules_java` results in an empty `java_import` rule being used.

Wyverald:
OK, this should be good to go. The remaining failures are already happening on master too.

PR #239, opened 2024-10-31
Copy over runfiles library from Bazel

fmeum:
The runfiles library can be maintained independently of Bazel releases and `bazel_tools` can refer to it via an alias.

Also set flags to build and test with a hermetic JDK 8 to ensure compatibility with that version.


fmeum:
@hvadehra @meteorcloudy I just realized that we could also move this over and delete it from bazel_tools. What do you think? We could fix https://github.com/bazelbuild/bazel/issues/24150 here.

comius:
The move sounds fine to me. I'll let @hvadehra do the review.

java/runfiles/src/main/java/com/google/devtools/build/runfiles/AutoBazelRepositoryProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+
+    String generatedClassName =
+        generatedClassPackage.isEmpty()
+            ? generatedClassSimpleName
+            : generatedClassPackage + "." + generatedClassSimpleName;
+
+    try (PrintWriter out =
+        new PrintWriter(
+            processingEnv.getFiler().createSourceFile(generatedClassName).openWriter())) {
+      out.printf("package %s;\n", generatedClassPackage);
```
fzakaria, line 98: This is the https://github.com/bazelbuild/bazel/issues/24150

fzakaria:
> @hvadehra @meteorcloudy I just realized that we could also move this over and delete it from bazel_tools. What do you think? We could fix [bazelbuild/bazel#24150](https://github.com/bazelbuild/bazel/issues/24150) here.

I included the fix in https://github.com/bazelbuild/bazel/pull/24161
It doesn't have any unit tests since there seem to be none to add.
A further contribution may be to include https://github.com/google/compile-testing at some point

fmeum:
Stacked on https://github.com/bazelbuild/rules_java/pull/238

fmeum:
@hvadehra This is ready for review now, CI is green.

fmeum:
@hvadehra Friendly ping

.bazelrc
```diff
@@ -1 +1,10 @@
 common --incompatible_disallow_empty_glob
+
+# Use hermetic JDKs for testing and ensure compatibliity with Java 8.
```
hvadehra, line 3: Are these needed for the rest of this change to work? If not, WDYT about a separate CI job with these flags?

test/repositories.bzl
```diff
@@ -0,0 +1,20 @@
+"""Test dependencies for rules_java."""
+
+load("@bazel_skylib//lib:modules.bzl", "modules")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
```
hvadehra, line 4: Any reason to not use http_jar from @rules_java itself? I think you said as much in https://github.com/bazelbuild/rules_java/pull/238

hvadehra (commented):
Sorry about the delay.

test/repositories.bzl
fmeum: Will update, that change just hadn't been merged yet and I didn't want to block this PR on it.

.bazelrc
fmeum: Without this, running `bazel test //...` locally would fail if the local JDK is more recent than the JDK in the Java compilation toolchain, which is very likely (I have JDK 23 locally). 

Do you see a downside to setting these flags? Ensuring that CI runs and local runs see the same result seems important for ease of contributing. This wasn't a problem prior to this change as the repo didn't actually contain any Java source files.

hvadehra: Yeah, that makes sense, but if it's not too much trouble, doing it in a separate change would be nicer.

fmeum: I split out https://github.com/bazelbuild/rules_java/pull/241

fmeum:
@Wyverald `http_jar` in `rules_java` doesn't work with Bazel 6 as it loads `cache.bzl`, which isn't available in that Bazel version.

test/repositories.bzl
fmeum: Added a comment to explain why.

Wyverald:
> @Wyverald `http_jar` in `rules_java` doesn't work with Bazel 6 as it loads `cache.bzl`, which isn't available in that Bazel version.

I'm aware of this (decided not to try to fix it since it's too much work). Is it blocking you?

fmeum:
If you are aware of it's fine, it just means that rules_java can't use it yet.

test/repositories.bzl
hvadehra: It should be possible to use via the compatibility proxy layer (redirecting to `@bazel_tools` instead for older Bazel). But there's no need to block this PR on that and can happen later.

hvadehra (dismissed):

hvadehra:
@fmeum This is causing a few issues on import, could you please change the newly added `BUILD` files to `BUILD.bazel`?

fmeum:
@hvadehra Done

fmeum:
@hvadehra Could you cut a new release so that this can be integrated into https://github.com/bazelbuild/bazel/pull/24219?

hvadehra:
Done: https://github.com/bazelbuild/bazel-central-registry/pull/3213

PR #243, opened 2024-11-25
Build bootclasspath in a UTF-8 environment

fmeum:
`java` and `javac` convert file and classpaths to absolute paths and thus require a UTF-8 locale to work under a path that contains non-ASCII characters.

Unblocks https://github.com/bazelbuild/bazel/pull/24457


fmeum:
@hvadehra

hvadehra:
An analysis test would be nice, if only to catch regressions. We can add an integration test for Bazel@HEAD after landing https://github.com/bazelbuild/bazel/pull/24457

fmeum:
@hvadehra I added one.

PR #272, opened 2025-02-06
Add make variables for runfiles location of $(JAVABASE) to support --no_legacy_external_runfiles.

thirtyseven:
Add make variables for runfiles locations of `$(JAVA)` and `$(JAVABASE)`. These are needed to access these paths when passing them to test rules via `env` when using `--no_legacy_external_runfiles`. The original make variables still work in exec contexts such as genrules.

### DIFF
--- a/java/common/rules/java_runtime.bzl
+++ b/java/common/rules/java_runtime.bzl
@@ -152,7 +152,8 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
-            "JAVABASE_RUNFILES": java_home_runfiles_path,
+            "JAVA_ROOTPATH": paths.join(java_home_runfiles_path, "bin/java"),
+            "JAVABASE_ROOTPATH": java_home_runfiles_path,
         }),
         ToolchainInfo(java_runtime = java_runtime_info),
     ]
--- a/java/common/rules/java_runtime.bzl
+++ b/java/common/rules/java_runtime.bzl
@@ -152,7 +152,7 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
-            "JAVA_ROOTPATH": paths.join(java_home_runfiles_path, "bin/java"),
+            "JAVA_ROOTPATH": java_binary_runfiles_path,
             "JAVABASE_ROOTPATH": java_home_runfiles_path,
         }),
         ToolchainInfo(java_runtime = java_runtime_info),
--- a/java/common/rules/java_runtime.bzl
+++ b/java/common/rules/java_runtime.bzl
@@ -152,6 +152,8 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
+            "JAVA_RUNFILES": java_binary_runfiles_path,
+            "JAVABASE_RUNFILES": java_home_runfiles_path,
         }),
         ToolchainInfo(java_runtime = java_runtime_info),
     ]
--- a/java/common/rules/java_runtime.bzl
+++ b/java/common/rules/java_runtime.bzl
@@ -152,7 +152,6 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
-            "JAVA_RUNFILES": java_binary_runfiles_path,
             "JAVABASE_RUNFILES": java_home_runfiles_path,
         }),
         ToolchainInfo(java_runtime = java_runtime_info),
--- a/toolchains/java_toolchain_alias.bzl
+++ b/toolchains/java_toolchain_alias.bzl
@@ -26,6 +26,8 @@ def _java_runtime_alias(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": str(toolchain.java_executable_exec_path),
             "JAVABASE": str(toolchain.java_home),
+            "JAVA_RUNFILES": str(toolchain.java_executable_runfiles_path),
+            "JAVABASE_RUNFILES": str(toolchain.java_home_runfiles_path),
         }),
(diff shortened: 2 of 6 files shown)


fmeum:
@meteorcloudy We might want to cherry-pick this into 8.1.0 as it makes it easier to adopt the `--nolegacy_external_runfiles` flip.

meteorcloudy:
@fmeum Do you mean bump the rules_java version in MODULE.tools to include this change? End users can also just depend on a newer rules_java to get this?

fmeum:
> @fmeum Do you mean bump the rules_java version in MODULE.tools to include this change? End users can also just depend on a newer rules_java to get this?

That is correct, but if we don't bump the minimum requirement, we can't document these Make variables on the versioned page for 8.1.0.

meteorcloudy:
OK, let's try to get this reviewed first, then let's see if we can make it for Bazel 8.1

@hvadehra Can you take a look?

toolchains/java_toolchain_alias.bzl
```diff
@@ -26,6 +26,8 @@ def _java_runtime_alias(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": str(toolchain.java_executable_exec_path),
             "JAVABASE": str(toolchain.java_home),
+            "JAVA_RUNFILES": str(toolchain.java_executable_runfiles_path),
```
hvadehra, line 29: I think we should instead be adding these to https://github.com/bazelbuild/rules_java/blob/af504cff01d441676f2d3e3b64c92d010f85d835/java/common/rules/java_runtime.bzl#L152 and then just forward that provider like we do below in this file: https://github.com/bazelbuild/rules_java/blob/af504cff01d441676f2d3e3b64c92d010f85d835/toolchains/java_toolchain_alias.bzl#L46

hvadehra (changes requested):
Please also add a test for this. Easiest might be to use this somehow in https://github.com/bazelbuild/rules_java/tree/master/test/repo

toolchains/java_toolchain_alias.bzl
thirtyseven: In order to make this change, I think I would have to make `java_runtime_alias` and `java_host_runtime_alias` identical by adding a dependency on `//toolchains:current_java_runtime` to `java_runtime_alias`. I can do this, but wanted to make sure it's your intent first.

hvadehra: My bad, please ignore the bit about just forwarding the provider. But I do think we need to make the same change also in [rules_java/java/common/rules/java_runtime.bzl](https://github.com/bazelbuild/rules_java/blob/af504cff01d441676f2d3e3b64c92d010f85d835/java/common/rules/java_runtime.bzl#L152)

thirtyseven: I have made the requested changes, however the test I added is failing due to `toolchain.java_executable_runfiles_path` appearing not to be a valid path. It's returning a path like:
```
/private/var/tmp/_bazel_thirtyseven/43339e4e51eaab2f90ecf6c85b4023b4/sandbox/darwin-sandbox/49/execroot/_main/../rules_java++toolchains+local_jdk/bin/java
```
which does not exist when running the test.

thirtyseven: I have removed `JAVA_RUNFILES` from this PR for now.

hvadehra:
Please fix the [buidifier failure](https://buildkite.com/bazel/rules-java-java/builds/2572#0194e68a-4403-4d84-99fb-cae3b2cd780b), the sh rules have been moved out of Bazel, so we'll need to load them from `@rules_sh`

toolchains/java_toolchain_alias.bzl
thirtyseven: Realized I could do something like:
```
"JAVA_RUNFILES": toolchain.java_home_runfiles_path + "/bin/java",
```
it's a little ugly but works. Thoughts?

hvadehra: Does that work? I would think the issue is that `JAVA_RUNFILES` is a special env var[^1] so if we'll want to use a different name.

[^1]: https://github.com/bazelbuild/bazel/blob/d796714abba0a5531edf795c53a3985300577303/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java#L82

thirtyseven: Not sure what a better name would be, I'm ok with leaving it out letting the user locate bin/java.

hvadehra (approved):

java/common/rules/java_runtime.bzl
```diff
@@ -152,6 +152,7 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
+            "JAVABASE_RUNFILES": java_home_runfiles_path,
```
fmeum, line 155: Not adding an equivalent for the `java` binary makes this difficult to use cross-platform (think Windows).

I don't think the deprecated env var `JAVA_RUNFILES` would necessarily conflict with a make var. It's something we could likely just get rid of at this point anyway.

But to avoid confusion, how about calling this `JAVA_ROOTPATH` or `JAVABASE_ROOTPATH` to match `$(rootpath ...)`?

fmeum (changes requested):

java/common/rules/java_runtime.bzl
thirtyseven: Sure, I've tried to address this.

java/common/rules/java_runtime.bzl
```diff
@@ -152,6 +152,8 @@ def _java_runtime_rule_impl(ctx):
         platform_common.TemplateVariableInfo({
             "JAVA": java_binary_exec_path,
             "JAVABASE": java_home,
+            "JAVA_ROOTPATH": paths.join(java_home_runfiles_path, "bin/java"),
```
fmeum, line 155: This won't work on Windows. Could you use `java_binary_runfiles_path` instead? If that somehow isn't the right path, you could also reuse the basename of the execpath.

thirtyseven: Done, `java_binary_runfiles_path` does actually work, the earlier issue I had must have been due to the `JAVA_RUNFILES` name conflict.

fmeum (approved):

hvadehra (approved):

PR #274, opened 2025-02-14
Add ppc64le platform support

Jenkins-J:
The platforms bazel module now defines a separate constraint for ppc6l4e. Utilize the new constraint value for the ppc64le architecture.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,7 +5,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "platforms", version = "0.0.11")
 bazel_dep(name = "rules_cc", version = "0.0.15")
 bazel_dep(name = "bazel_features", version = "1.11.0")
 bazel_dep(name = "bazel_skylib", version = "1.6.1")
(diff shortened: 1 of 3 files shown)


hvadehra:
@Jenkins-J Please fix CI. I think it should just be a matter of updating the `@platforms` dep in the `WORKSPACE` files as well: https://github.com/bazelbuild/rules_java/blob/4b91b00e003fe856dd0c98ee64f5aa034dcb91e2/java/rules_java_deps.bzl#L159

hvadehra:
Thanks. 

So, the existing ppc jdks were already for ppc64le but before this change we were using the less specific `ppc` cpu constraint from `@platforms`. So I think what we want to do is replace all references to `ppc` with `ppc64le` (instead of adding it additionally) and drop these hacky lines: 

 - https://github.com/bazelbuild/rules_java/blob/4e83620b038575b4f9e20f3dc581072035c081a0/java/bazel/repositories_util.bzl#L100-L101
 - https://github.com/bazelbuild/rules_java/blob/4e83620b038575b4f9e20f3dc581072035c081a0/java/bazel/repositories_util.bzl#L129-L130

Please also sync to HEAD to resolve the merge conflict.

hvadehra (approved):

PR #276, opened 2025-02-24
Set use_default_shell_env = True consistently.

benjaminp:
About half the action registrations in rules_java formerly set this. Be consistent.

### DIFF
--- a/java/common/rules/android_lint.bzl
+++ b/java/common/rules/android_lint.bzl
@@ -136,6 +136,7 @@ def _android_lint_action(ctx, source_files, source_jars, compilation_info):
         tools = tools,
         arguments = args_list,
         execution_requirements = {"supports-workers": "1"},
+        use_default_shell_env = True,
     )
     return android_lint_out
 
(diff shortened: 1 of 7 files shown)


fmeum:
It's unfortunately still common for users to use `--action_env` to set variables for repo rules, which when combined with `use_default_shell_env = True` results in cache invalidation. Do you see a use case for `--action_env` for hermetic JDKs? If not, then setting the default based on whether the `JAVA_HOME` path is absolute could be a safer choice.

benjaminp:
I'm trying to pass certain environmental variables to every hermetic JVM invocation, so `use_default_shell_env = True` only for local jdks wouldn't suffice.

What would also be useful is the ability to configure `--action_env` by mnemonic or perhaps platform.

fmeum:
> What would also be useful is the ability to configure --action_env by mnemonic or perhaps platform.

Yes, that would be really useful and much less of a footgun.

crt-31:
Actually, I came here to argue that we should always set use_default_shell_env to **False**. 
Setting it to true causes cache misses, and rules_java should be oriented to hermetic by default.

PR #277, opened 2025-02-26
Make `local_java_repository` a repository rule

meteorcloudy:
Try to addresse: https://github.com/bazelbuild/bazel/pull/25356#issuecomment-2675083765

### DIFF
--- a/toolchains/local_java_repository.bzl
+++ b/toolchains/local_java_repository.bzl
@@ -317,7 +317,7 @@ Args:
   version: optionally java version
   **kwargs: additional arguments for repository rule
 """
-local_java_repository_rule = repository_rule(
+local_java_repository = repository_rule(
     implementation = _local_java_repository_impl,
     local = True,
     configure = True,
--- a/toolchains/local_java_repository.bzl
+++ b/toolchains/local_java_repository.bzl
@@ -286,7 +286,38 @@ def _create_auto_config_error_build_file(repository_ctx, *, local_java_runtime_n
         False,
     )
 
-_local_java_repository_rule = repository_rule(
+"""Defines runtime and compile toolchains for a local JDK.
+
+Register the toolchains defined by this macro as follows (where `<name>` is the value of the
+`name` parameter):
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@fmeum Can you take a look why this change is breaking the presubmit? Maybe related to https://cs.opensource.google/bazel/bazel/+/8715e9ae6f5c2fd42607018b3adaad14bf601a1c?

fmeum:
@meteorcloudy You need to augment the rule definition to use `name` as the default for `runtime_name`.

meteorcloudy:
Oh, sorry for missing that!

hvadehra:
I'm not sure we can get easily rid of the macro. Things I thought could work:
 - Use a computed default for `runtime_name` -> will still see the wrong value of name (i.e. `toolchains+<name>` instead of just `<name>`. Even if that were not the case, `attr.string()` current does not accept a `StarlarkFunction`.
 - Use an initializer -> not implemented for repository rules

IIRC the `runtime_name` attribute was actually introduced to avoid splitting on `~` or `+` when that delimiter was changed. The only way I see for this change to work is to go back to splitting. But we'd need to conditionally split based on what Bazel version we're on (or by heuristically checking for either char I suppose).

Alternatively, does it work to keep the macro and make the repo rule public? Or are we now disallowing macros from being used in this fashion?

fmeum:
We can use the new `repository_ctx.original_name` attr if available (Bazel 8.1.1 and later).

hvadehra:
> 
We can use the new repository_ctx.original_name attr if available (Bazel 8.1.1 and later).

TIL. How do we keep this working for Bazel 6 and 7 though?

fmeum:
> > We can use the new repository_ctx.original_name attr if available (Bazel 8.1.1 and later).
> 
> TIL. How do we keep this working for Bazel 6 and 7 though?

We could export the repo rule and use `repository_ctx.original_name` in it if available, but keep exporting the macro wrapper around it. At some point we can replace the macro with an alias for the repo rule.

Issue #275, opened 2025-02-15
Add better support for generated resources


joca-bt:
[`java_library.srcs`](https://bazel.build/reference/be/java#java_library.srcs) supports `.srcjar` files to facilitate including generated sources. `.properties` files included in `java_library.srcs` are filtered and included in `java_library.resources`. Consider extending the current behaviour so that `.properties` files included in `.srcjar` files are also included in `java_library.resources`. This will facilitate the usage of generated resources.

For example, I have a `genrule` that generates `.java` and `.properties` files and I don't necessarily know all of them during the analysis phase for declaration purposes. For this reason, my rule instead generates a `.srcjar` file. However, the `.properties` files inside that `.srcjar` file are being skipped.

There are other alternative ways of going about it: 

- There's already an internal [`resource_jars`](https://github.com/bazelbuild/rules_java/blob/8.9.0/java/common/rules/impl/basic_java_library_impl.bzl#L95) field that could perhaps be reused for this effect (at the moment it's always provided as empty)? However, straight including `.srcjar` files in this field won't achieve the desired outcome as this can result in `.java` files appearing inside the final `.jar` file.

- Support something like `.resourcejar` files in `java_library.resources`. This would be a breaking change :pensive:.

hvadehra:
You could also have your genrule generate two outputs, a `foo.srcjar` with just the `.java` sources, and a `foo-props.jar` with the `.properties` files. 

```
genrule(
  name = "gen", 
  outs = ["foo.srcjar", "foo-props.jar"],
  cmd = 
    """
    # do whatever to generate the files
    zip -r $(location foo.srcjar) . -i *.java
    zip -r $(location foo-props.jar) . -i *.properties
    """, 
)

java_import(name = "props", jars = [":foo-props.jar"])

java_library(
  name = "lib",
  srcs = [":foo.srcjar"],
  runtime_deps = [":props"],
)
```

which should achieve what you want.

hvadehra:
https://github.com/bazelbuild/rules_java/blob/3d757492427c10c73e48cbd0aa0623f12bfb769a/java/java_single_jar.bzl#L76 can help with that.

joca-bt:
`java_single_jar` is missing from [`java/defs.bzl`](https://github.com/bazelbuild/rules_java/blob/master/java/defs.bzl). Should it be added there?

hvadehra:
[...]

PR #278, opened 2025-03-03
Add java_single_jar to defs.bzl

joca-bt:


joca-bt:
Partially addresses https://github.com/bazelbuild/rules_java/issues/275#issuecomment-2693938854

hvadehra:
The `defs.bzl` pattern is considered legacy, prefer to load the specific bzl file for the symbols you need. Using a single bzl file to export all symbols will cause unnecessary loading / fetching. For instance, in this file, one may end up fetching @protobuf even if one doesn't use any of the proto rules.

joca-bt:
The documentation still recommends providing defs.bzl https://bazel.build/rules/deploying.

hvadehra:
The latest guidance I'm aware of is https://docs.google.com/document/d/1L1JFgjpZ7SrBinb24DC_5nTIELeYDacikcme-YcA7xs/edit?usp=sharing

The page you linked probably need to be updated, which is actually called out as a `TODO` in the doc :)

joca-bt:
Ok, thanks! Feel free to close the issue if you want.

PR #291, opened 2025-04-24
Increase maxwarns to -1

fzakaria:
Please see https://github.com/bazelbuild/bazel/issues/25927#issuecomment-2825206105

It's important not to filter out any warnings otherwise the lint warnings are not surfaced and then any attempt to do `-Werror` can fail with what may be observed as spurious (affect other parts of the codebase).

This is maybe a "breaking change" only if someone has enabled `-Werror` and may now see additional errors -- although that was in fact their intention...

### DIFF
--- a/toolchains/default_java_toolchain.bzl
+++ b/toolchains/default_java_toolchain.bzl
@@ -49,10 +49,6 @@ BASE_JDK9_JVM_OPTS = [
     # makes it go to stderr instead.
     "-Xlog:disable",
     "-Xlog:all=warning:stderr:uptime,level,tags",
-
-    # Please see https://github.com/bazelbuild/bazel/issues/25927#issuecomment-2825206105
-    # we need to make sure not to filter any warnings so that Bazel can trigger them as errors.
-    "-Xmaxwarns -1",
 ]
 
 JDK9_JVM_OPTS = BASE_JDK9_JVM_OPTS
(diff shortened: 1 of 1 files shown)


hvadehra:
The motivation for this sounds reasonable to me, but I think we should only be setting defaults that make sense for everyone, all the time. So, it's probably best if we actually trim down the default list to an empty one rather than expanding it (so the default javac behavior is what everyone gets out-of-the-box). If you'd like to have different behavior, consider registering a custom java_toolchain with those options.

cushon:
> The motivation for this sounds reasonable to me, but I think we should only be setting defaults that make sense for everyone, all the time. So, it's probably best if we actually trim down the default list to an empty one rather than expanding it (so the default javac behavior is what everyone gets out-of-the-box). If you'd like to have different behavior, consider registering a custom java_toolchain with those options.

FWIW I think this default sort of makes sense for everyone, the status quo is that errors may not be shown if there are too many warnings, and then there's no way to tell why the build failed from the log. There is an underlying bug causing that which should be fixed, but until then this default is helpful for everyone. And I also agree in theory it would be nice to use javac defaults, but in practice I'm not sure there's a strong reason to change this default upstream, the issue here is Bazel-specific.

fzakaria:
Yea -- basically without this PR users of rules_java are subject to potential bugs that I've documented in the bug.
If someone (@cushon ) can recommend where to try and patch it upstream  -- I can look into that as well; but for now this is a pretty reasonable change.

hvadehra:
> the issue here is Bazel-specific.

Is this because JavaBuilder does not propagate the `-Werror` to the compilation task? If so, ISTM we should perhaps not propagate `-Xmaxwarns` either (passing -1 instead), and do the limiting on the JavaBuilder side (after filtering ignored diagnostics). Is that right?

PR #292, opened 2025-05-04
Configure coverage helpers for the test exec group

fmeum:
These tools run on the test action's execution platform and should thus be configured for it.


fmeum:
@katre

PR #297, opened 2025-06-02
Make `java_toolchain` usable with `genrule`

fmeum:
`genrule` accepts toolchain types since https://github.com/bazelbuild/bazel/commit/0e876b1794d4db58b72949014401d22cc65d94a1, but there is no way to get an exec-configured Java runtime in this way since `java_runtime` toolchains typically match on the target rather than the exec platform.

Instead, the `java_toolchain` used for compilation is made usable for this purpose by forwarding the `DefaultInfo` and `TemplateVariableInfo` of its `java_runtime`.

### DIFF
--- a/java/common/rules/java_toolchain.bzl
+++ b/java/common/rules/java_toolchain.bzl
@@ -139,7 +139,14 @@ def _java_toolchain_impl(ctx):
         _timezone_data = ctx.file.timezone_data,
     )
     toolchain_info = ToolchainInfo(java = java_toolchain_info)
-    return [java_toolchain_info, toolchain_info, DefaultInfo()]
+    providers = [java_toolchain_info, toolchain_info]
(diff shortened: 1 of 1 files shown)


fmeum:
@hvadehra @katre Could you take a look? CI failures are caused by changes to Bazel@HEAD and should not be related to this PR.

katre (approved):

cushon:
> The "alias" targets, which also exist in Bazel, are decent workarounds. Their major downside is that they won't really make sense in a future with proper toolchain support on genrules. From the perspective of a Bazel 9 user, having to put `toolchains = ["@rules_java//toolchains:current_host_java_runtime"]` on a genrule would seem pretty inconsistent compared to just using the toolchain type as for other rulesets.
>

PR #301, opened 2025-06-11
Add support for `--incompatible_compact_repo_mapping_manifest`

fmeum:
Adds support for the new `_repo_mapping` format introduced in https://github.com/bazelbuild/bazel/issues/26262 to the runfiles library.


meteorcloudy:
@fmeum Can you sync to HEAD, for the CI breakage, it probably needs https://github.com/fmeum/rules_java/commit/fecb9fbe5e5417997bb8dda24c1873a130baf7d4

meteorcloudy (approved):
Thanks! Should we explicitly point out that the new runfiles library is also compatible with the old format?

PR #312, opened 2025-08-06
Add a module extension for registering local/remote jdks

hvadehra:


toolchains/extensions.bzl
```diff
@@ -0,0 +1,53 @@
...
+                remote.version,
+                target_compatible_with = remote.target_compatible_with,
+                prefix = remote.prefix,
+                sha256 = remote.sha256,
+                strip_prefix = remote.strip_prefix,
+                urls = remote.urls,
+            )
+
+_local = tag_class(attrs = {
+    "name": attr.string(mandatory = True),
```
fmeum, line 30: Module extensions typically don't work well if they accept user-provided names while also collecting tags from transitive deps. There just is no good way to handle collisions.

Since this extension would only wrap simple repos, could `use_repo_rule` be an alternative? If not because the repo rules really are macros, what's blocking them from being rules?

hvadehra: Thanks for the heads up. This is just me experimenting for https://github.com/bazelbuild/bazel/issues/26589

If we do go ahead with this, I was imagining only permitting root modules to use this. WDYT?

PR #357, opened 2026-05-18
Add loongarch64 support

zhaixiaojuan:


toolchains/BUILD
```diff
@@ -144,6 +144,7 @@ cc_library(
         "@bazel_tools//src/conditions:linux_mips64": [":jni_md_header-linux"],
         "@bazel_tools//src/conditions:linux_ppc64le": [":jni_md_header-linux"],
         "@bazel_tools//src/conditions:linux_riscv64": [":jni_md_header-linux"],
+        "@bazel_tools//src/conditions:linux_loongarch64": [":jni_md_header-linux"],
```
hvadehra: this doesn't exist in https://github.com/bazelbuild/bazel/blob/master/src/conditions/BUILD.tools

However, https://github.com/bazelbuild/bazel/commit/1c29dff364fd34d7e6158c812cfd6d96f66be747 has since landed so we can resolve the TODO from L137 and simplify all this. Could you please do that instead?

hvadehra (changes requested):

toolchains/BUILD
zhaixiaojuan: > 
> However, [bazelbuild/bazel@1c29dff](https://github.com/bazelbuild/bazel/commit/1c29dff364fd34d7e6158c812cfd6d96f66be747) has since landed so we can resolve the TODO from L137 and simplify all this. Could you please do that instead?

Okay, I will try to do that.

hvadehra (approved):

PR #363, opened 2026-06-30
Propagate `.jdeps` for `java_import`

### LINKED ISSUE #362: classpath reduction defeated by transitive `java_import` dependencies
I was playing with Bazel's `--experimental_java_classpath=bazel_no_fallback` flag and noticed classpath reduction was defeated by transitive `java_import` dependencies.

For example:
```
$ curl --remote-name-all -L 'https://repo1.maven.org/maven2/com/google/guava/guava/33.6.0-jre/guava-33.6.0-jre.jar'  'https://repo1.maven.org/maven2/com/google/guava/failureaccess/1.0.3/failureaccess-1.0.3.jar'
$ cat MODULE.bazel
bazel_dep(name = "rules_java", version = "9.6.1")
$ cat X.java
void main() {
  var _ = com.google.common.util.concurrent.SettableFuture.create();
}
$ cat BUILD.bazel
load("@rules_java//java:java_binary.bzl", "java_binary")
load("@rules_java//java:java_import.bzl", "java_import")

java_binary(
    name = "X",
    srcs = ["X.java"],
    main_class = "X",
    deps = [":guava"],
)

java_import(
    name = "guava",
    jars = ["guava-33.6.0-jre.jar"],
    deps = [":failureaccess"],
)

java_import(
    name = "failureaccess",
    jars = ["failureaccess-1.0.3.jar"],
)
$ bazel build --experimental_java_classpath=bazel_no_fallback --java_language_version=25 --java_runtime_version=25 //:X
INFO: Analyzed target //:X (0 packages loaded, 0 targets configured).
ERROR: BUILD:4:12: Building X.jar (1 source file) failed: (Exit 1): java failed: error executing Javac command (from java_binary rule target //:X) external/rules_java++toolchains+remotejdk25_linux/bin/java '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED' '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED' ... (remaining 19 arguments skipped)
X.java:2: error: cannot access InternalFutureFailureAccess
  var _ = com.google.common.util.concurrent.SettableFuture.create();
                                                          ^
  class file for com.google.common.util.concurrent.internal.InternalFutureFailureAccess not found
Target //:X failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.215s, Critical Path: 0.12s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```
Is this expected? This behavior can easily turn classpath reduction into a significant pessimization on a real codebase.

fmeum:
Downstream Java compilation actions need accurate `.jdeps` files for their reduced classpath optimization. With this change, `java_import` now always runs `ImportDepsChecker` to generate this file, but possibly suppresses its outputs. Note that Bazel (but not Blaze) always silences the output.

Fixes #362

### DIFF
--- a/test/java/common/rules/java_import_tests.bzl
+++ b/test/java/common/rules/java_import_tests.bzl
@@ -178,6 +178,59 @@ def _test_deps_impl(env, targets):
         "{package}/depjar.jar",
     ])
 
+# Regression test for https://github.com/bazelbuild/rules_java/issues/362: an import with deps
+# records a jdeps proto as its compile_jdeps so that reduced classpaths of downstream compilations
+# are not missing its transitive dependencies.
+def _test_compile_jdeps_propagated_for_deps(name):
+    util.helper_target(
+        java_import,
+        name = name + "/import-jar",
+        jars = ["import.jar"],
+        deps = [name + "/depjar"],
+    )
+    util.helper_target(
+        java_import,
+        name = name + "/incomplete-jar",
+        jars = ["incomplete.jar"],
+        deps = [name + "/depjar"],
+        tags = ["incomplete-deps"],
+    )
+    util.helper_target(
+        java_import,
+        name = name + "/depjar",
+        jars = ["depjar.jar"],
+    )
+
+    analysis_test(
+        name = name,
+        impl = _test_compile_jdeps_propagated_for_deps_impl,
+        targets = {
+            "with_deps": name + "/import-jar",
+            "without_deps": name + "/depjar",
+        },
+        # Requires the rules_java Starlark implementation to be used.
+        attr_values = {"tags": ["min_bazel_9"]},
+    )
+
+def _test_compile_jdeps_propagated_for_deps_impl(env, targets):
+    with_deps = [
+        f.basename
+        for f in targets.with_deps[JavaInfo]._compile_time_java_dependencies.to_list()
+    ]
+    env.expect.that_collection(with_deps).contains_exactly(["jdeps.proto"])
+
+    incomplete_deps = [
+        f.basename
+        for f in targets.without_deps[JavaInfo]._compile_time_java_dependencies.to_list()
+    ]
+    env.expect.that_collection(incomplete_deps).contains_exactly(["jdeps.proto"])
+
+    without_deps = [
+        f.basename
+        for f in targets.without_deps[JavaInfo]._compile_time_java_dependencies.to_list()
+    ]
+    env.expect.that_collection(without_deps).contains_exactly(["jdeps.proto"])
+
 # Regression test for b/262751943.
 def _test_commandline_contains_target_label(name):
     util.helper_target(
@@ -959,6 +1012,7 @@ def java_import_tests(name):
             _test_simple,
             _test_with_java_library,
             _test_deps,
+            _test_compile_jdeps_propagated_for_deps,
             _test_commandline_contains_target_label,
             _test_java_library_allows_import_in_deps,
             _test_module_flags,
--- a/test/java/common/rules/java_import_tests.bzl
+++ b/test/java/common/rules/java_import_tests.bzl
@@ -178,7 +178,7 @@ def _test_deps_impl(env, targets):
         "{package}/depjar.jar",
     ])
 
-# Regression test for https://github.com/bazelbuild/rules_java/issues/362: an import with deps
+# Regression test for https://github.com/bazelbuild/rules_java/issues/362: every java_import
 # records a jdeps proto as its compile_jdeps so that reduced classpaths of downstream compilations
 # are not missing its transitive dependencies.
 def _test_compile_jdeps_propagated_for_deps(name):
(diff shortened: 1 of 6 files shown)


toolchains/default_java_toolchain.bzl
```diff
@@ -76,6 +76,7 @@ _DEFAULT_JAVA_LANGUAGE_VERSION = "11"
 
 # Default java_toolchain parameters
 _BASE_TOOLCHAIN_CONFIGURATION = dict(
+    deps_checker = Label("@remote_java_tools//:ImportDepsChecker"),
```
fmeum, line 79: This is a breaking change for Bazel as the hardcoded mode is `error`. Can I add a new attribute that makes it configurable on the toolchain level? I can set the default to `silence` to match the current behavior or to `error` as a breaking change with a simple migration.

Are you precompiling the tool to a native image at Google?

fmeum:
@cushon

toolchains/default_java_toolchain.bzl
cushon: > Are you precompiling the tool to a native image at Google?

It's a prebuilt jar that runs on the JVM, but it's also a validation action that is not on the critical path (https://github.com/bazelbuild/rules_java/issues/362#issuecomment-4842458374).

Using a native image seems like a good idea if it's going to run as a non-validation action so the `.jdeps` can be used for reduced classpaths.

cushon: > Can I add a new attribute that makes it configurable on the toolchain level?

Bazel supports `tags = ["incomplete-deps"]` on `java_import`. Having a real attribute (on the toolchain and/or individual `java_import`s) is probably better than leaning on tags for this.

cc @hvadehra

fmeum: My main concern is that external repos using `java_import` would be very difficult to opt out unless there is a global (toolchain) switch.

hvadehra: This should only be a breaking change for Bazel 7, which is okay I guess. I also don't see why we need a separate switch - one can just unset `deps_checker` on the toolchain.

test/java/common/rules/java_import_tests.bzl
```diff
@@ -178,6 +178,59 @@ def _test_deps_impl(env, targets):
...
+        jars = ["import.jar"],
+        deps = [name + "/depjar"],
+    )
+    util.helper_target(
+        java_import,
+        name = name + "/incomplete-jar",
+        jars = ["incomplete.jar"],
+        deps = [name + "/depjar"],
+        tags = ["incomplete-deps"],
+    )
```
hvadehra, line 197: unused?

test/java/common/rules/java_import_tests.bzl
```diff
@@ -178,6 +178,59 @@ def _test_deps_impl(env, targets):
...
+
+    analysis_test(
+        name = name,
+        impl = _test_compile_jdeps_propagated_for_deps_impl,
+        targets = {
+            "with_deps": name + "/import-jar",
+            "without_deps": name + "/depjar",
+        },
+        # Requires the rules_java Starlark implementation to be used.
+        attr_values = {"tags": ["min_bazel_9"]},
```
hvadehra: starlark impl is used from Bazel 8

test/java/common/rules/java_import_tests.bzl
```diff
@@ -178,6 +178,59 @@ def _test_deps_impl(env, targets):
...
+    )
+
+def _test_compile_jdeps_propagated_for_deps_impl(env, targets):
+    with_deps = [
+        f.basename
+        for f in targets.with_deps[JavaInfo]._compile_time_java_dependencies.to_list()
+    ]
+    env.expect.that_collection(with_deps).contains_exactly(["jdeps.proto"])
```
hvadehra: written this way, the test will have terrible failure messages if/when it fails. use `java_info_subject.from_target()` instead.

(same below)

test/java/common/rules/java_import_tests.bzl
```diff
@@ -178,6 +178,59 @@ def _test_deps_impl(env, targets):
...
+
+    incomplete_deps = [
+        f.basename
+        for f in targets.without_deps[JavaInfo]._compile_time_java_dependencies.to_list()
```
hvadehra: this is not the `incomplete_deps` target, is that unintentional?

test/java/common/rules/java_import_tests.bzl
```diff
@@ -959,6 +1012,7 @@ def java_import_tests(name):
             _test_simple,
             _test_with_java_library,
             _test_deps,
+            _test_compile_jdeps_propagated_for_deps,
```
hvadehra, line 1004: I think tests for the `ImportDepsChecker` action would be useful as well - particularly around the value of `--checking_mode`

hvadehra (changes requested):

test/java/common/rules/java_import_tests.bzl
fmeum: I added a test and realized that Bazel always forces the `silence` mode. Which is good for backwards compatibility, I guess, but probably worth making configurable at some point.

test/java/common/rules/java_import_tests.bzl
```diff
@@ -178,6 +178,77 @@ def _test_deps_impl(env, targets):
...
+        target = name + "/import-jar",
+        # The rules_java Starlark implementation is used from Bazel 8 on.
+        attr_values = {"tags": ["min_bazel_8"]},
+    )
+
+def _test_import_deps_checker_checking_mode_impl(env, target):
+    # java_import only generates the jdeps proto, it never fails the build on incomplete deps.
+    assert_action = env.expect.that_target(target).action_named("ImportDepsChecker")
+    assert_action.contains_flag_values([
+        ("--checking_mode", "silence"),
```
hvadehra: This will fail internally (where this will be "error"). Mind putting just this test case in `test/java/bazel/rules/java_import_tests.bzl`?

hvadehra (changes requested):

test/java/common/rules/java_import_tests.bzl
hvadehra: Thanks!

test/java/common/rules/java_import_tests.bzl
fmeum: Done

hvadehra (approved):

hvadehra:
Just a heads up: adding the jdep to `JavaInfo` breaks things internally with downstream kotlin rules. Will take some time to figure out.

PR #365, opened 2026-07-09
Mark remote JDK toolchain config repositories reproducible

dzbarsky:
The `_toolchain_config` repository rule creates the remote JDK `*_toolchain_config_repo` repositories from deterministic `WORKSPACE` and `BUILD.bazel` content. Return `repo_metadata(reproducible = True)` so Bazel can use the repository contents cache for these repositories, with a `hasattr` guard for Bazel versions that do not provide `repository_ctx.repo_metadata`.

Validation:

- Bazel 9.1.1 query of all 200 targets in the 40 remote JDK toolchain config repositories
- Bazel 7.6.1 query of `@remotejdk21_macos_aarch64_toolchain_config_repo//:all`
- `buildifier -mode=check toolchains/remote_java_repository.bzl`

PR #367, opened 2026-07-09
Return repo_metadata from repository_rules

keith:
This is used by bazel to determine that these rules are remotely
cacheable with the remote repo contents cache.

### DIFF
--- a/toolchains/local_java_repository.bzl
+++ b/toolchains/local_java_repository.bzl
@@ -176,6 +176,10 @@ def _local_java_repository_impl(repository_ctx):
       repository_ctx: repository context
     """
 
+    repository_metadata = None
+    if hasattr(repository_ctx, "repo_metadata"):
+        repository_metadata = repository_ctx.repo_metadata(reproducible = True)
+
     java_home = _determine_java_home(repository_ctx)
 
     local_java_runtime_name = repository_ctx.attr.runtime_name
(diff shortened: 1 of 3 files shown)


hvadehra:
https://github.com/bazelbuild/rules_java/pull/365 was opened first but this one seems more complete. 

So I'd appreciate it if you and @dzbarsky can decide on which to proceed with.

toolchains/local_java_repository.bzl
```diff
@@ -176,6 +176,10 @@ def _local_java_repository_impl(repository_ctx):
       repository_ctx: repository context
     """
 
+    repository_metadata = None
+    if hasattr(repository_ctx, "repo_metadata"):
+        repository_metadata = repository_ctx.repo_metadata(reproducible = True)
```
hvadehra, line 181: Is a local java repository necessarily reproducible? 

We detect the java_home from environment variables if not explicitly set. Even with an explicit `java_home`, what if it points to a symlink (that can change under us)? 

Does it even make sense to cache a local repo like this remotely?

dzbarsky: I believe a repo that has a symlink to a absolute on-host location will get cached in the local repo contents cache but not the remote one (but @fmeum can confirm!)

I do think it would be safer/more obviously correct not to cache this one, but I'd also hope that it's completely lazy and that we don't force/autoregister the autodetecting toolchain for this language the way rules_cc and others do, though I haven't verified :/

dzbarsky:
> #365 was opened first but this one seems more complete.
> 
> So I'd appreciate it if you and @dzbarsky can decide on which to proceed with.

I don't mind closing mine if we get this one in quickly, if we think discussion of which repos to cache might drag on, then it might make sense to take the obviously correct ones upfront

toolchains/local_java_repository.bzl
fmeum: It should generally be safe to mark such repos reproducible since there "contents" are just the symlink to an absolute path, where the path is determines by attributes. 

Whether both caches behave correctly in this case is not something I have checked yet, but if not, that just means they are buggy.

hvadehra (approved):

PR #368, opened 2026-07-24
Support RunfilesGroupInfo

malt3:
Teach the Java rules to describe their runfiles as named, ordered groups so that downstream packaging rules can build more efficient artifacts.

`java_binary`, `java_test`, `java_library`, and `java_import` now return `RunfilesGroupInfo` (and `RunfilesGroupMetadataInfo`) from the `rules_runfiles_group` ruleset alongside `DefaultInfo`. Instead of seeing a single flat runfiles tree, a packaging rule (e.g. a container-image or archive rule) can split a binary's runfiles into layers and order them so that the content that changes least often lands in the most cacheable layers:

  * the JDK / java_runtime at the foundation tier (never merged),
  * `java_import` targets (typically third-party jars) at the shared-deps tier,
  * first-party `java_library` code, the binary's own jars, and the executable at the executable tier.

Libraries propagate fine-grained per-target groups up through their dependents; the binary collects them, adds its own groups, and tags every group it produces with the `rules_java` merge affinity so that JVM-shaped groups stay together when a packager has to merge groups to fit a layer limit. `java_import` gains a `runfiles_weight` attribute so dependency-management rulesets can hint at relative sizes to guide those merge decisions.

Consumers that don't understand `RunfilesGroupInfo` are unaffected and keep using `DefaultInfo.default_runfiles`.

Adds a dependency on `rules_runfiles_group` for both Bzlmod and WORKSPACE setups.

## Validation

I took the liberty to host a [BCR mirror that has support enabled for rules_java and rules_img](https://bazel-contrib.github.io/rules_runfiles_group/). I also created a [demo repo to show the effect](https://github.com/malt3/runfiles_group_demo).

This shows two `java_binary` targets with overlapping dependencies. Both are packaged as container images.
Without `RunfilesGroupInfo`, the whole `java_binary` is stored as a single layer and there is no sharing (only the base image layer is shared):

With `RunfilesGroupInfo`, individual runfiles from `java_library` targets get their own layers and most of the content is shared across the container images:

### DIFF
--- a/java/bazel/rules/bazel_java_binary.bzl
+++ b/java/bazel/rules/bazel_java_binary.bzl
@@ -16,6 +16,8 @@
 load("@bazel_features//:features.bzl", "bazel_features")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@rules_cc//cc:find_cc_toolchain.bzl", "use_cc_toolchain")
+load("@rules_runfiles_group//runfiles_group:lib.bzl", "runfiles_groups")
+load("@rules_runfiles_group//runfiles_group:providers.bzl", "RunfilesGroupInfo")
 load("//java/common:java_semantics.bzl", "semantics")
 load(
     "//java/common/rules:android_lint.bzl",
@@ -26,6 +28,16 @@ load("//java/common/rules:rule_util.bzl", "merge_attrs")
 load("//java/common/rules/impl:java_binary_deploy_jar.bzl", "create_deploy_archives")
 load("//java/common/rules/impl:java_binary_impl.bzl", "basic_java_binary", "binary_provider_helper")
 load("//java/common/rules/impl:java_helper.bzl", "helper")
+load(
+    "//java/common/rules/impl:runfiles_group_support.bzl",
+    "BINARY_GROUP",
+    "JAVA_RUNTIME_GROUP",
+    "MERGE_AFFINITY",
+    "RUNFILES_GROUP_ATTRS",
+    "collect_entries",
+    "own_kind",
+    "runfiles_groups_enabled",
+)
 load("//java/private:java_info.bzl", "JavaInfo")
 
 def _bazel_java_binary_impl(ctx):
@@ -100,9 +112,11 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
 
     runfiles = default_info.runfiles
 
+    toolchain_files = depset()
     if executable:
         runtime_toolchain = semantics.find_java_runtime_toolchain(ctx)
-        runfiles = runfiles.merge(ctx.runfiles(transitive_files = runtime_toolchain.files))
+        toolchain_files = runtime_toolchain.files
+        runfiles = runfiles.merge(ctx.runfiles(transitive_files = toolchain_files))
 
     test_support = helper.get_test_support(ctx)
     if test_support:
@@ -113,6 +127,8 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
         runfiles = runfiles,
         executable = default_info.executable,
     )
+
+    _add_runfiles_group_provider(providers, ctx, java_attrs, executable, toolchain_files, test_support)
 
     info = providers.pop("InternalDeployJarInfo")
     create_deploy_archives(
@@ -127,6 +143,52 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
     )
 
     return providers.values()
+
+def _add_runfiles_group_provider(providers, ctx, java_attrs, executable, toolchain_files, test_support):
+    if not runfiles_groups_enabled(ctx):
+        return
+
+    own = [runfiles_groups.entry(
+        name = ctx.label,
+        content = java_attrs.runtime_jars,
+        kind = own_kind(ctx.label),
+        rank = runfiles_groups.RANK_EXECUTABLE,
+        merge_affinity = MERGE_AFFINITY,
+    )]
+
+    executable_group = None
+    if executable:
+        # toolchain_files is only populated for a target with an executable.
+        own.append(runfiles_groups.entry(
+            name = JAVA_RUNTIME_GROUP,
+            content = toolchain_files,
+            kind = "foundation",
+            rank = runfiles_groups.RANK_FOUNDATION,
+            do_not_merge = True,
+            merge_affinity = MERGE_AFFINITY,
+        ))
+        own.append(runfiles_groups.entry(
+            name = BINARY_GROUP,
+            content = depset([executable]),
+            kind = own_kind(ctx.label),
+            rank = runfiles_groups.RANK_EXECUTABLE,
+            merge_affinity = MERGE_AFFINITY,
+        ))
+        executable_group = BINARY_GROUP
+
+    deps = [ctx.attr.deps, ctx.attr.runtime_deps]
+    if test_support:
+        deps.append(test_support)
+
+    providers["RunfilesGroupInfo"] = RunfilesGroupInfo(
+        entries = collect_entries(
+            ctx,
+            deps = deps,
+            data = [getattr(ctx.attr, "data", [])],
+            own = own,
+        ),
+        executable_group = executable_group,
+    )
 
 def _get_coverage_runner(ctx):
     if ctx.configuration.coverage_enabled and ctx.attr.create_executable:
@@ -370,6 +432,7 @@ logic as the Java package of source files. For example, a source file at
             executable = True,
         ),
     } if not bazel_features.rules._has_launcher_maker_toolchain else {},
+    RUNFILES_GROUP_ATTRS,
 )
 
 def make_java_binary(executable):
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -9,6 +9,7 @@ bazel_dep(name = "platforms", version = "0.0.11")
 bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "bazel_features", version = "1.30.0")
 bazel_dep(name = "bazel_skylib", version = "1.6.1")
+bazel_dep(name = "rules_runfiles_group", version = "0.1.0")
 bazel_dep(name = "protobuf", version = "32.1", repo_name = "com_google_protobuf")
 bazel_dep(name = "zlib", version = "1.3.1.bcr.5")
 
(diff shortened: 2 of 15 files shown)


java/bazel/rules/bazel_java_binary.bzl
```diff
@@ -128,6 +134,34 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
...
+    if test_support:
+        all_deps.append(test_support)
+    dep_collected = lib.collect_groups(ctx, all_deps)
+    data_collected = lib.collect_groups(ctx, ctx.attr.data) if hasattr(ctx.attr, "data") else struct(groups = {}, metadata = None)
+
+    groups = {}
+    groups.update(dep_collected.groups)
+    groups.update(data_collected.groups)
+
+    own_group_name = str(ctx.label)
```
fmeum: This has high memory overhead for targets in external repos. Could the group name be a `Label` instead so that no new string needs to be retained?

java/bazel/rules/bazel_java_binary.bzl
```diff
@@ -128,6 +134,34 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
...
+    groups[own_group_name] = ctx.runfiles(transitive_files = java_attrs.runtime_jars)
```
fmeum: A runfiles object has a much higher retained heap cost than the depset of files it wraps. Would it be possible to retain depsets directly as an optimization when none of the other fields of `runfiles` are needed?

java/bazel/rules/bazel_java_binary.bzl
malt3: The data structure for this is a dict. I don't think dicts can use `Label` as a key. When trying your suggestion I'm getting:

```
Error: keywords must be strings, not Label
```

Happy to modify the data structure for lower memory usage if there's a way!

java/bazel/rules/bazel_java_binary.bzl
malt3: Yes. I think it's straightforward to do this, since consumers of RGI are expected to use helpers from the `lib` struct in rules_runfiles_group.
What's the best way to check if a value is a runfiles object vs. a depset?

malt3:
I discussed with @fmeum that it's possible to make this provider a lot more memory efficient. I'll also add a global flag to enable/disable the emition of this provider for memory conscious users.
I'll update this PR accordingly.

java/bazel/rules/bazel_java_binary.bzl
malt3: This is now implemented

java/bazel/rules/bazel_java_binary.bzl
malt3: This is now implemented

java/bazel/rules/bazel_java_binary.bzl
```diff
@@ -128,6 +143,52 @@ def bazel_base_binary_impl(ctx, is_test_rule_class):
...
     return providers.values()
 
+def _add_runfiles_group_provider(providers, ctx, java_attrs, executable, toolchain_files, test_support):
+    if not runfiles_groups_enabled(ctx):
+        return
+
+    own = [runfiles_groups.entry(
+        name = ctx.label,
+        content = java_attrs.runtime_jars,
+        kind = "first_party",
```
fmeum: Should this be gated on the current module? Not all Bazel-built Java targets would be necessarily be considered first-party (e.g. the protobuf runtime). Changing this later wouldn't be difficult though, so that's certainly not a blocker from my point of view.

fmeum (approved):

java/bazel/rules/bazel_java_binary.bzl
malt3: Let's gate it on the module. I like it.

fmeum: Perhaps the repo is both simpler and more accurate. A `use_repo_rule` for `http_archive` would show up as being owned by your module, but more likely than not is third-party source.

malt3: Done. I'm using the repo name to check this.

fmeum (approved):
You can mark this rfr.

malt3:
> You can mark this rfr.

I will, but it's referencing a prerelease version of `rules_runfiles_group`.
I'll kick off a new release of that module now. Just FYI for any reviewer.

EDIT: done. Now pointing at `rules_runfiles_group@0.1.0`

hvadehra:
Instead of making changes to various rules repos, have you considered achieving this with an aspect? ISTM this is the exactly kind of thing aspects were designed for.

malt3:
@hvadehra I have considered it, but I haven't found a good way to do this so far.
The biggest blocker I see is that currently, the Aspect has to be tailor-made for every language ruleset.
I don't know of an efficient way to walk the deps of a `java_binary` (or other language rule) and decide for each dep which runfiles that dep has contributed to the runfiles of the leaf.

Or to make this more concrete: In the implementation function of an aspect, I can inspect the `DefaultInfo` of the current target, but I cannot get an incremental view of said `DefaultInfo` (given a target `x` that inherits `DefaultInfo`/runfiles from `a` and `b` and merges them into it's own runfiles, get the subset of files added by `x` that were not provided by `a` or `b`).
The only way that I know of is to call `to_list()` on the depsets and figure out what was added by a specific target. This is undesirable IMO.

If we figure out a way to do the same work efficiently in an Aspect, I'd be happy to do so.

hvadehra:
I could imagine having a `runfiles_group_support.bzl` in `rules_java` with something like:

```
def process_target_from_aspect(ctx):
   if not _is_rules_java_target(ctx):
     return None
   
   if ctx.rule.kind == "java_binary":
     // do what this PR currently does and return the right RunfilesGroupInfo()
   elif ...
   
```

with tests ensuring the rule implementations and the function stay in sync (at least in the ways we care about).

malt3:
@hvadehra I like your proposal. Let me rephrase it in my own words to ensure we are on the same page:

- `java_library` targets do not provide `RunfilesGroupInfo` directly
- `java_binary` would apply an aspect to it's own deps that is specific to rules_java
- The Aspect collects information on runfiles groups
- `java_binary` **does** return `RunfilesGroupInfo` (if the global configuration setting demands it)

Can you confirm if I understand your idea correctly? If we agree on that, I can implement your proposal.

hvadehra:
What I was imagining was:

 - Your ruleset (rules_img or rules_runfiles_group) defines the aspect. This aspect propagates along all attrs. In it's impl function, it can call `@rules_java`'s support function described above to get `RunfilesGroupInfo` for a `rules_java` target.
 - `rules_img` can attach the aspect on `image_from_binary`'s `binary` attribute.

malt3:
I like this idea in principle, but I found a problem when trying to build a working proof-of-concept:

Any ruleset that wants to add support (rules_java, rules_python, etc..) needs to define a support function.
How does the aspect implementation function defined in `@rules_runfiles_group` call the correct support function?

>  In it's impl function, it can call @rules_java's support function described above to get RunfilesGroupInfo for a rules_java target.

I'd like to avoid the need for rules_runfiles_group to depend on every module that wants to offer a helper function (as rules_java would).

One idea I have is to instead return a single instance of a provider that holds the callback function. That's a bit weird as well, but it avoids a dep on `rules_java`.
Do you have any alternatives in mind?

malt3:
@fmeum had the following idea when I asked him how we might be able to implement a callback mechanism without explicit dependencies (translated from German):

> Hi Malte. Off the top of my head, I could imagine the following:
> 
> rules_runfiles_group offers a provider whose only field is such a callback function, and specifies a well-known attribute.
> rules_lang then defines a target that returns this callback via this provider.
> All rules in rules_lang have this well-known implicit attribute pointing at that one target.
> If that works, it would be only constant overhead and you wouldn't need a dependency on rulesets.

@hvadehra could you elaborate on your idea? Would that work for you?

It would look something like this:

```starlark
# //mylang:runfiles_group_support.bzl
load("@rules_runfiles_group//runfiles_group:callback.bzl", "RunfilesGroupCallbackInfo")
load("@rules_runfiles_group//runfiles_group:lib.bzl", "runfiles_groups")
load("@rules_runfiles_group//runfiles_group:providers.bzl", "RunfilesGroupInfo")

def _library_groups(target, ctx, _payload):
    # every language ruleset can implement this as they wish.
    # The rules_java implementation looks different from this example.
    return RunfilesGroupInfo(entries = runfiles_groups.collect(
        ctx,
        deps = [ctx.rule.attr.deps, ctx.rule.attr.exports],
        data = [ctx.rule.attr.data],
        own = [runfiles_groups.entry(
            name = target.label,
            content = depset(target[MyLangInfo].own_files),
            kind = "first_party",
            merge_affinity = "mylang",
        )],
    ))

def _callback_impl(_ctx):
    return [RunfilesGroupCallbackInfo(describe = _library_groups)]

mylang_library_callback = rule(implementation = _callback_impl)
```

```starlark
# //mylang:rules.bzl -- loads nothing from @rules_runfiles_group.
mylang_library = rule(
    implementation = _mylang_library_impl,
    attrs = dict(_MYLANG_LIBRARY_ATTRS, **{
        # The well-known attribute signals support for runfiles group info.
        # The function within this attr can be called in the aspect implementation function.
        "_runfiles_group_callback": attr.label(default = Label("//mylang:library_callback")),
    }),
)
```

The well-known attribute name `_runfiles_group_callback` would allow `rules_runfiles_group` to detect support.

hvadehra:
Apologies for the radio silence, got busy with other things.

The idea of using a provider/attribute for the callback is interesting. It's still tighter coupling than I was hoping for. But if you don't want a `rules_runfiles_group` -> `rules_X` dep, we might not be able to do any better today.

I'd like to spend a few days thinking about whether there are enhancements to the Bazel rules API that could help (my intuition is any solution here might benefit many other rule extension FRs). If I don't get anywhere, lets proceed with this PR next week.

fmeum:
A limitation I didn't consider is that an aspect's propagation predicate can't see the target behind a label attribute and thus wouldn't be able to see the provider. If the aspect needs to make propagation decisions based on rules_java's structure, more "well-known attributes" would be needed.

malt3:
One idea I had: `rules_runfiles_group` could provide a module extension that rules_java (and other modules) could call in their MODULE.bazel to register callbacks that the aspect could then invoke.

hvadehra:
@malt3 lets go ahead with this. The well-defined attribute idea SGTM.

PR #375, opened 2026-08-16
Silence protobuf sun.misc.Unsafe warnings from the direct header comp…

### LINKED ISSUE #374: `header_compiler_direct` (TurbineDirect) emits protobuf `sun.misc.Unsafe` warnings on JDK 24+ that no toolchain option can silence
### Description

When header compilation runs on a **JDK 24+** runtime, every direct-header compilation prints:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by com.google.protobuf.UnsafeUtil
         (.../java_tools/turbine_direct_binary_deploy.jar)
WARNING: Please consider reporting this to the maintainers of class com.google.protobuf.UnsafeUtil
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
```

TurbineDirect bundles protobuf, whose `UnsafeUtil` calls the terminally deprecated `sun.misc.Unsafe` memory-access methods; on JDK 24+ (JEP 498) the JVM warns by default (mode `warn`). Root cause of the Unsafe usage is protocolbuffers/protobuf#20760.

### Why it can't be silenced today

The obvious fix — `--sun-misc-unsafe-memory-access=allow` — cannot be routed to the TurbineDirect JVM from a toolchain:

- **`turbine_jvm_opts`** is passed only to the Turbine *worker*
  (`header_compiler`), not to `header_compiler_direct`
  (`java/common/rules/java_toolchain.bzl`: `header_compiler` gets
  `turbine_jvm_opts`; `header_compiler_direct` gets a separately-computed
  `header_compiler_direct_jvm_opts` that is hardcoded to
  `["-Dturbine.ctSymPath=…"]`).
- **`--jvmopt`** applies to `java_binary`/`java_test` execution, not to
  header-compilation actions (verified with `aquery`: the flag does not appear
  in the TurbineDirect command line).
- `header_compiler_direct_jvm_opts` is **not exposed** as a `java_toolchain` /
  `default_java_toolchain` attribute, and `header_compiler_direct` cannot be
  dropped (rules_java `fail()`s when header compilation is requested without it).

So a downstream project on JDK 24+ (e.g. Gerrit Code Review) has no supported way to quiet the warning.

Bazel itself tracks protobuf#20760 per-tool (its root `BUILD` adds the flag to `turbine_jvm_opts`;

`import_deps_checker` adds it to `jvm_flags`; commit `a121ec75d2` narrowed a global `--jvmopt` down to `import_deps_checker`), but none of those cover the `header_compiler_direct` header-compilation path — so the warning surfaces for every rules_java consumer building Java on JDK 24+.

### Reproduction

Any `java_library` built with a toolchain whose `java_runtime` is JDK 24+.

`bazel aquery 'mnemonic("Turbine", //your:lib)'` shows the TurbineDirect command carries only `-Dturbine.ctSymPath=…` and a fixed set of `--add-opens` — no way to add `--sun-misc-unsafe-memory-access=allow`.

### Proposed fix (see the PR)

Add `--sun-misc-unsafe-memory-access=allow` to `header_compiler_direct_jvm_opts` in `_java_toolchain_impl`, gated on the runtime feature version (the flag is rejected on JDK ≤ 22, so it must be version-guarded; `java_runtime.version` is `0` when unknown):

```python
if java_runtime and java_runtime.version >= 24:
    header_compiler_direct_jvm_opts = header_compiler_direct_jvm_opts + [
        "--sun-misc-unsafe-memory-access=allow",
    ]
```

Verified with `aquery` against a JDK-25 toolchain (the flag now appears in the TurbineDirect command) and a JDK-21 toolchain (the flag is correctly omitted, so the JVM still starts). This fixes it once for all consumers instead of requiring each project to work around it — and the obvious per-project workaround (`turbine_jvm_opts`) does not even reach the direct header compiler.

### Environment

- rules_java 9.8.0 (bundled `remote_java_tools` TurbineDirect)
- JDK 24+ runtime (e.g. `remotejdk_25`)
- Related: protocolbuffers/protobuf#20760 (root cause), JEP 498

davido:
…iler

turbine_direct (header_compiler_direct) bundles protobuf, whose UnsafeUtil calls terminally-deprecated sun.misc.Unsafe methods. On JDK 24+ (JEP 498) the JVM prints a warning for each such call during header compilation:

  WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
  WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by
  com.google.protobuf.UnsafeUtil (.../java_tools/turbine_direct_binary_deploy.jar)

Unlike the Turbine worker, header_compiler_direct does not receive turbine_jvm_opts, and --jvmopt does not reach header-compilation actions, so a toolchain currently has no way to silence these warnings for the direct header compiler. Add --sun-misc-unsafe-memory-access=allow to header_compiler_direct_jvm_opts, gated on the runtime feature version since the flag is rejected on JDK <= 22.

See https://github.com/protocolbuffers/protobuf/issues/20760.

Closes #374

### DIFF
--- a/java/common/rules/java_toolchain.bzl
+++ b/java/common/rules/java_toolchain.bzl
@@ -91,6 +91,9 @@ def _java_toolchain_impl(ctx):
     else:
         header_compiler_direct_data = []
         header_compiler_direct_jvm_opts = []
+
+    header_compiler_direct_jvm_opts = header_compiler_direct_jvm_opts + ctx.attr.header_compiler_direct_opts
+
     if ctx.attr.oneversion_allowlist and ctx.attr.oneversion_whitelist:
         fail("oneversion_allowlist and oneversion_whitelist are mutually exclusive")
     oneversion_allowlist = ctx.file.oneversion_allowlist if ctx.file.oneversion_allowlist else ctx.file.oneversion_whitelist
@@ -586,6 +589,18 @@ Labels of data available for label-expansion in turbine_jvm_opts.
         "turbine_jvm_opts": attr.string_list(
             doc = """
 The list of arguments for the JVM when invoking turbine.
+            """,
+        ),
+        "header_compiler_direct_opts": attr.string_list(
+            doc = """
+Additional options passed to the direct header compiler, verbatim and without
+location or make-variable expansion.
+
+Warning: these are only meaningful when the direct header compiler runs on a
+JVM, such as TurbineDirect's deploy jar. A native direct header compiler such
+as turbine_direct_graal parses them as Turbine command-line options, so do not
+set JVM options (for example --sun-misc-unsafe-memory-access=allow) when a
+native direct header compiler is in use.
             """,
         ),
         "xlint": attr.string_list(


java/common/rules/java_toolchain.bzl
```diff
@@ -91,6 +91,23 @@ def _java_toolchain_impl(ctx):
...
+    # silenced for the direct header compiler. Do not pass this JVM-only
+    # flag to turbine_direct_graal: its native binary parses these values
+    # as Turbine command-line options.
+    if (
+        java_runtime and
+        java_runtime.version >= 24 and
+        _is_java_launcher_tool(ctx.attr.header_compiler_direct)
+    ):
+        header_compiler_direct_jvm_opts = header_compiler_direct_jvm_opts + [
+            "--sun-misc-unsafe-memory-access=allow",
```
hvadehra: rather than hardcoding this here, I would prefer we do one of the following:

1. Reuse `ctx.attr.turbine_jvm_opts` if `header_compiler_direct` is the same as `header_compiler` i.e. the Turbine deploy jar
2. Add a `turbine_direct_jvm_opts` attribute

hvadehra (changes requested):

java/common/rules/java_toolchain.bzl
davido: Done.

java/common/rules/java_toolchain.bzl
```diff
@@ -91,6 +91,22 @@ def _java_toolchain_impl(ctx):
     else:
         header_compiler_direct_data = []
         header_compiler_direct_jvm_opts = []
+
+    turbine_direct_jvm_opts = get_internal_java_common().expand_java_opts(
```
hvadehra: Lets not complicate things unnecessarily. Just use the opts as is without expansion for now.

java/common/rules/java_toolchain.bzl
```diff
@@ -91,6 +91,22 @@ def _java_toolchain_impl(ctx):
...
+        ctx,
+        "turbine_direct_jvm_opts",
+        tokenize = False,
+        exec_paths = True,
+    )
+    if turbine_direct_jvm_opts and not _is_deploy_jar_tool(ctx.attr.header_compiler_direct):
```
hvadehra: If we're going with a dedicated attribute, I don't think we need to perform all this validation. Lets just pass the user-specified opts as is.

Maybe just a warning note in the attribute docs about graal vs deploy jar.

java/common/rules/java_toolchain.bzl
```diff
@@ -588,6 +610,14 @@ Labels of data available for label-expansion in turbine_jvm_opts.
 The list of arguments for the JVM when invoking turbine.
             """,
         ),
+        "turbine_direct_jvm_opts": attr.string_list(
```
hvadehra: For consistency lets call this `header_compiler_direct_opts`

toolchains/default_java_toolchain.bzl
```diff
@@ -70,6 +70,22 @@ DEFAULT_JAVACOPTS = [
     "-Xmaxwarns -1",
 ]
 
+TURBINE_DIRECT_JAVA_FALLBACK_JVM_OPTS = select({
```
hvadehra: I don't think this makes sense - the `header_compiler_direct` is a public attribute and user-settable. Expecting the non-graal executable based on platform isn't correct.

hvadehra (changes requested):

toolchains/default_java_toolchain.bzl
davido: Done — dropped the select and reverted default_java_toolchain.bzl entirely, so it's out of the diff. The attribute is now purely opt-in: the default toolchain no longer sets it, and toolchains that use a JVM direct header compiler pass the flag themselves.

java/common/rules/java_toolchain.bzl
davido: Done.

java/common/rules/java_toolchain.bzl
davido: Done — removed the guard entirely. Replaced it with a warning note in the attribute docs about graal vs deploy jar, as you suggested.

java/common/rules/java_toolchain.bzl
davido: Done — passed through verbatim now, no expansion. Documented that in the attr doc too.

hvadehra (approved):

PR #377, opened 2026-08-20
Apply execution requirements only to Java test runs

realityforge:
Using tags for requirements affects every action that builds a test. Add a java_test attribute backed by testing.ExecutionInfo so callers can scope requirements such as network access to the runner.

Keep the automatic requires-darwin requirement and cover both direct requirements and merge behavior with analysis tests.


fmeum:
Have you tried using `exec_properties = {"test.some-tag": ""}`? That applies only to the test action itself.

We should probably unflip `--incompatible_allow_tags_propagation` though.

realityforge:
Ah - yes thank you @fmeum - you are totally right. Just another feature of Bazel that I have yet to learn. Thanks for the heads up!

PR #381, opened 2026-09-10
Add jni_md_header for NetBSD

zakinko:
The :jni target pulls the machine-dependent jni_md.h out of the JDK per
host, with a java_runtime_files target and a select arm for each
platform.  FreeBSD and OpenBSD are there; add NetBSD, whose JDK ships
include/netbsd/jni_md.h just as the others do, so that Bazel's native
code builds against the right header when it is bootstrapped there.

The config_setting is defined here rather than taken from
@bazel_tools//src/conditions, so that it resolves with every Bazel
version rather than only those carrying bazelbuild/bazel#31069.
@platforms//os:netbsd has been in the platforms module since 0.0.11,
which this module already depends on.

DragonFly BSD needs @platforms//os:dragonfly, which is not in a
platforms release yet (bazelbuild/platforms#143), so it will follow in a
separate change.


hvadehra:
Unless we cherrypick https://github.com/bazelbuild/bazel/pull/31069 to all supported Bazel versions, this will break everyone using this target with a `rules_java` version that includes this change.

If you'd like to get this in sooner, we should to update to a `platforms` release containing https://github.com/bazelbuild/platforms/pull/143 and define the `config_setting`s locally.

zakinko:
Right -- dropped the @bazel_tools//src/conditions reference.  The
config_setting is defined in toolchains/BUILD now, on
@platforms//os:netbsd, which has been in platforms since 0.0.11 and so
resolves with any Bazel version; nothing here depends on
bazelbuild/bazel#31069 landing.

DragonFly is out of this PR for now, since @platforms//os:dragonfly
needs bazelbuild/platforms#143 to be released first.  I'll send it
separately once that is in.

hvadehra (approved):

PR #382, opened 2026-09-15
Make `proguard_allowlister` optional and drop the @bazel_tools default

fmeum:
The Java toolchain's `proguard_allowlister` defaulted to `@bazel_tools//tools/jdk:proguard_whitelister`, a Python script that only lives in `@bazel_tools` because the native Java and Android rules used it. Keeping it there forces `@bazel_tools` to depend on rules_python, and it is being removed together with the other leftovers of the native rules (see the ongoing cleanup of the embedded tools in bazelbuild/bazel).

With this change the attribute is optional. When a toolchain doesn't set it, `proguard_specs` are forwarded as they are instead of being copied to `validated_proguard/...`. Toolchains that want the check can point the attribute at their own allowlister; the affected tests do that through a test-only toolchain with a stand-in allowlister, selected via `--extra_toolchains`.

rules_android's toolchain has the same default and will get a corresponding change.


hvadehra:
If this is just for the bazel_tools -> rules_python dep, see https://github.com/bazelbuild/bazel/commit/c95f2fcee298b179b4f4643f80c5efa1182ad6d9

PR #383, opened 2026-09-23
Fix native library path separator on Windows

gharrma:
On Windows, java.library.path is split by ";" instead of ":".

Regression test added.

Note for Googlers: this fix is in service of b/149162146#comment19


gharrma:
I had to disable the added test on Bazel 7.x since that version of Bazel still uses the built-in Java rules. After that, CI passes and the change is now ready for review.

PR #384, opened 2026-09-24
Starlarkify collect_native_deps_dirs()

gharrma:
This code previously called JavaStarlarkCommon.collectNativeLibsDirs() from Bazel, but we can implement this in Starlark instead.

As a side effect, this change avoids a Windows-specific bug in Bazel's implementation of collectNativeLibsDirs(). If a library provides both an interface library (.lib) and a dynamic library (.dll), Bazel would end up using neither, and the library would never end up on java.library.path. I would have liked to add a regression test for this Windows-specific bug, but there is a related rules_cc bug that makes it tricky: https://github.com/bazelbuild/rules_cc/pull/899. I plan to add a Windows test later once all the fixes are in.

Note for Googlers: this fix is in service of b/149162146#comment19.


gharrma:
`test_java_binary_native_library_path_includes_transitive_deps` initially passed on buildkite, but then failed in google3, which surprised me. I uploaded a new commit with a guess for what will fix it. I'll monitor TAP presubmit tomorrow to see if it works.

